PR #6, opened 2021-10-29
Product images

### LINKED ISSUE #4: Build Docker images for all products
## Description

We need to Dockerize all our products.
These Docker images will probably evolve over time but initially we need them for all our supported versions, they need to include JMX where relevant and any authorizers needed.

We decided to keep the Dockerfiles in one central location (i.e. this repository) and not colocate them with the operators.

We would like to use the RedHat UBI images as a starting point (they do exist with JDKs as well).

One thing I've always been unsure about is a tagging strategy (e.g. when to use "latest") so that should be part of this.
Our "old" `tar.gz` packages could serve as a starting point but they were created manually and this might be a good time to make this workflow repeatable and automated.

Please also investigate "buildah" as a replacement for Docker.

### Questions
Q: Should we add the packages to our Nexus before building the images ?
A: ??? (It would simplify the image pipeline a lot. For example different Nifi versions have different download urls)

Q: Kafka - which Scala version 2.12, 1.13 or both ?
A: Images are now build with both and also tagged with the scala version.

Q: Trino - should the images also contain the trino cli and trino jdbc ?
A: ???

Q: Tagging with latest is not possible with the current tagging scheme. See meeting notes on versions and tags below. Is this a problem ?
A: ??? (operators need to know exact versions when they create pods anyway)

### Docker images

- [x] OPA
- [x] Apache ZooKeeper
- [x] Apache NiFi
- [x] Apache Kafka
- [x] Trino
- [x] Apache Hadoop
- [x] Apache Spark
- [x] Apache HBase
- [x] Apache Hive
- [x] Apache Superset
- [x] (Maybe) Apache Druid - Coordinate with Felix

### Operators adapted

moved to -> https://github.com/stackabletech/issues/issues/136

### Docs updated (thx Felix)

- [ ] OPA
- [ ] Apache ZooKeeper
- [ ] Apache NiFi
- [ ] Apache Kafka
- [ ] Trino
- [ ] Apache Hadoop
- [ ] Apache Spark
- [ ] Apache HBase
- [ ] Apache Hive
- [ ] Apache Superset
- [ ] Apache Druid

razvan:
See open questions in the linked ticket!

Fixes #4 

Create application Dockerfiles and GH actions to build them.


maltesander (approved):
LGTM.

Issue #25, opened 2021-11-30
Reduce size of Docker images


siegfriedweber:
The Docker images are much larger than the official ones:
```
$ docker images | grep trino
docker.stackable.tech/stackable/trino        362-stackable0                 4a6cc6f9f9dd   2 weeks ago     4GB
trinodb/trino                                362                            8c4c3cc8b6b5   2 months ago    1.33GB
```

The size of the images should be decreased to speed up the download. This would also make the integration tests faster and decrease the chance of hitting timeouts.

One possible optimization is to not rename the installation directory in a separate layer:
```
RUN mkdir /stackable/trino-server && \
    mv /stackable/trino-server-${PRODUCT}/* /stackable/trino-server && \
    rm -rf /stackable/trino-server-${PRODUCT}
```

lfrancke:
See also https://github.com/stackabletech/issues/issues/149

_This_ ticket is only about fixing the immediate issue of having Docker images that are over twice the size they need to be.
There are more improvements to be made in the CI pipeline later but this is not the ticket for hat.

PR #26, opened 2021-12-01
Reduce image sizes

maltesander:
fixes https://github.com/stackabletech/docker-images/issues/25

### DIFF
--- a/kafka/Dockerfile
+++ b/kafka/Dockerfile
@@ -20,7 +20,7 @@ RUN groupadd -r stackable --gid=1000 && \
 USER stackable
 
 ARG PRODUCT
-ARG SCALA=2.13
+ARG SCALA
 ARG OPA_AUTHORIZER
 
 WORKDIR /stackable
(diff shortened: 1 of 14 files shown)


kafka/Dockerfile
```diff
@@ -9,30 +12,26 @@ RUN microdnf update && \
...
+COPY stackable /stackable
+RUN groupadd -r stackable --gid=1000 && \
+    useradd -r -g stackable --uid=1000 stackable && \
+    chown -R stackable:stackable /stackable
 
 USER stackable
 
 ARG PRODUCT
-ARG SCALA
+ARG SCALA=2.13
```
razvan, line 23: I recommend to not hard code the scala version here but set it in conf.py.

razvan (approved):
ok

kafka/Dockerfile
teozkr: @razvan This was just a default that would have been overridden by the matrix build anyway, as far as I can tell. Is there an actual reason to drop Scala 2.12 support here?

teozkr: Not saying I'm particularly pro-2.12, just that I'd like for that to be an explicit decision taken somewhere rather than just a consequence of misunderstandings.. :P

maltesander: yeah it would have been overridden by the build. I thought the default makes it easier to build manually. Im good with both (keeping / leaving it out)

PR #27, opened 2021-12-02
Added update_config script to nifi (For NiFi 1.15)

fhennig:
Since NiFi 1.14 the sensitive props key is a mandatory property. It is also a secret, so it is now mounted into an enviroment variable. This script takes the value from the variable and writes it to the config.


fhennig:
This is related to https://github.com/stackabletech/nifi-operator/pull/125

maltesander (approved):
LGTM. Do not forget to run the action to push to our repo!

fhennig:
will do

PR #29, opened 2021-12-14
Adds script to remove vulnerable log4shell .class files from products we roll out.

soenkeliebau:


soenkeliebau:
I've added this to the architecture agenda for today to dicsuss if this is sufficient for what we want to do or if we want to do more.
More would mostly be thinking about implementing this: https://github.com/corretto/hotpatch-for-apache-log4j2
Would need to be per containern though.

druid/Dockerfile
```diff
@@ -32,8 +32,26 @@ RUN curl -L https://repo.stackable.tech/repository/packages/druid/apache-druid-$
...
+# TODO: This can be restricted to target only versions which do not honor the environment
+#   varible that has been set above but this has not currently been implemented
+COPY shared/log4shell.sh /bin
+RUN /bin/log4shell.sh /stackable/druid-${PRODUCT}
+
+# Ensure no vulnerable files are left over
+# This will currently report vulnerable files being present, as it also alerts on
+# SocketNode.class, which we do not remove with our scripts.
+# Further investigation will be needed whether this should also be removed.
+COPY shared/log4shell_1.1.2-log4shell_Linux_x86_64 /bin/log4shell_scanner
```
razvan, line 52: The file `shared/log4shell_1.1.2-log4shell_Linux_x86_64` is missing.

druid/Dockerfile
```diff
@@ -32,8 +32,26 @@ RUN curl -L https://repo.stackable.tech/repository/packages/druid/apache-druid-$
...
+# This variable is supported as of Log4j version 2.10 and
+# disables the vulnerable feature
 ENV LOG4J_FORMAT_MSG_NO_LOOKUPS=true
 
+# For earlier versions this script removes the .class file that contains the
+# vulnerable code.
+# TODO: This can be restricted to target only versions which do not honor the environment
+#   varible that has been set above but this has not currently been implemented
+COPY shared/log4shell.sh /bin
+RUN /bin/log4shell.sh /stackable/druid-${PRODUCT}
```
razvan, line 46: There is no `/stackable/druid-${PRODUCT}` folder. Remove the `${PRODUCT}` suffix.

```suggestion
RUN /bin/log4shell.sh /stackable/druid
```

siegfriedweber:
This is a permanent solution for a temporary problem. If we are really affected then we have to fix it, if not then not.

druid/Dockerfile
soenkeliebau: I've added the 'apache-' prefix, I'd rather not rely on the symlink, just to exclude any potentially weird behavior of the script when ran with symlinks.

druid/Dockerfile
soenkeliebau: added

soenkeliebau:
> This is a permanent solution for a temporary problem. If we are really affected then we have to fix it, if not then not.

Not sure if I follow - are you saying we should tackle this in a completely different way?

My reasoning for this was that we do not really have the manpower to go and fix this upstream, which is the place where it needs fixing, so to avoid shipping a self-compiled / patched version of the products this is an external step that we can add for now to be on the safe side.

As and when upstream projects are fixed we can work to remove this, or rather run it only for affected versions.

druid/Dockerfile
razvan: Sorry I didn't mention but the path needs to be updated in all Dockerfiles.

soenkeliebau: I've gone over the files once more, I only found mismatches for Hive & Spark though..

razvan: I see this when I try to build the Zookeeper image:

razvan (commented):
needs work

druid/Dockerfile
teozkr: Looks like the zookeeper path should be `/stackable/apache-zookeeper-${PRODUCT}-bin` rather than `/stackable/zookeeper-${PRODUCT}`

razvan (approved):
lgtm

PR #31, opened 2021-12-28
The write-myid script is no longer required.

maltesander:
The myid is written via an extra container in the zookeeper operator (using id of the statefulset pod(s)).


teozkr (approved):
Do we want to move the new ID writer into the image at some point? Doesn't have to be now, though..

maltesander:
> Do we want to move the new ID writer into the image at some point? Doesn't have to be now, though..

Im not sure. I actually like to have the control over that via the operators and not the dependency on the images.

PR #34, opened 2022-01-20
Feature/tools image

razvan:
Tools image with `curl` and in the future maybe other utilities. used by init containers for various checks.


maltesander (approved):
LGTM.

PR #36, opened 2022-01-21
Airflow image

adwk67:
- Airflow image for use with the airflow operator
- Uses a standard apache/airflow image with default versions (airflow + python) which can be overridden with the build script
- Defaults to airflow 2.2.3 and python 3.8 (some airflow components require at least python 3.7)
- Build tested locally with docker and helm


siegfriedweber (approved):
LGTM

PR #37, opened 2022-01-24
Adds OpenSSL and bcrypt to nifi image

soenkeliebau:
During the init phase of NiFi containers two additional tools are required to prepare the envionment:
- OpenSSL to prepare the keystore used for TLS
- bcrypt which is used to encrypt the admin password from a secret and insert it into the config file

As bcrypt tool I have chosen https://github.com/bitnami/bcrypt-cli mostly because it does not require python, node or any other heavyweight dependency.

I contemplated writing our own tool in Rust which would have been something like 50 lines, but saw no real benefit in this.

### DIFF
--- a/nifi/Dockerfile
+++ b/nifi/Dockerfile
@@ -7,8 +7,13 @@ SHELL ["/bin/bash", "-o", "pipefail", "-c"]
 RUN microdnf update && \
     microdnf install java-11-openjdk-headless --nodocs && \
     microdnf install tar gzip zip && \
-    microdnf install shadow-utils && \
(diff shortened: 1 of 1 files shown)


razvan (approved):
Works but I want to mention that we have a "tools" image for one-shot tasks.

soenkeliebau:
While that would work for the init container and openssl, it wouldn't for the bcrypt tool, because this is needed inside of the actual container that runs NiFi, since any in-place changes to the Configs are ephemeral across container restarts.

soenkeliebau:
closed in favor of #38

PR #38, opened 2022-01-24
Adds installation of openssl and bcrypt to tools image.

soenkeliebau:
During the init phase of NiFi pods two additional tools are required to prepare the envionment:
- OpenSSL to prepare the keystore used for TLS
- bcrypt which is used to encrypt the admin password from a secret and insert it into the config file

As bcrypt tool I have chosen https://github.com/bitnami/bcrypt-cli mostly because it does not require python, node or any other heavyweight dependency.

I contemplated writing our own tool in Rust which would have been something like 50 lines, but saw no real benefit in this.


razvan (approved):
lgtm

PR #39, opened 2022-01-25
Enable HDFS metrics for all services

razvan:
Added journal node configuration for the jmx exporter and set a delay of 10 seconds before exporting the metrics.

The metrics are now also used for health checks in K8S and therefore added a delay of 10 seconds to allow service containers to start up properly.


teozkr:
What's the purpose of the delay?

razvan:
> What's the purpose of the delay?

To ensure that the RPC service thread is properly initialized. Without it, the namenodes fail to start randomly.

snocke (approved):

PR #48, opened 2022-02-09
Parameterize the base image tag of the Superset dockerfile

### LINKED ISSUE #35: Superset version is hard coded in the Dockerfile
The Superset version is hard coded in the Dockerfile but the version from the configuration file `conf.py` should be taken.

Superset is similar to AirFlow because both are based on the same Python framework and their images can be built in the same way. The Dockerfile for AirFlow already uses a parameterized approach which should also be used in the Dockerfile for Superset.

siegfriedweber:
The base image tag for the Superset dockerfile is now parameterized and must be set in the configuration file for every Superset version.

The Stackable Superset image uses the official Superset image as base. The official images are not tagged properly. They use commit hashes as tags. The base image tag is not a dependency (like a Python or Scala version), so it should not be appended to the image tag. This is solved by prefixing the version item containing the base image tag with an underscore. Version items prefixed with an underscore are now only available as build arguments in the Dockerfile but not recognized as dependencies.

Closes #35

### DIFF
--- a/build_product_images.py
+++ b/build_product_images.py
@@ -1,4 +1,4 @@
-#!/bin/env python
+#!/usr/bin/env python
 """
 Build and possibly publish product images. It doesn't login to any registry when publishing
 but it assumes a `docker login` has been performed before.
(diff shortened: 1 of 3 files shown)


razvan (approved):
Very smart! Seems a bit abusive and a bit too implicit but it works. As a quick-fix I like it.

PR #50, opened 2022-02-14
Enable statsd logging in Superset

siegfriedweber:
Enable statsd logging in Superset.

This is needed for stackabletech/superset-operator#128.

### DIFF
--- a/superset/stackable/superset_config.py
+++ b/superset/stackable/superset_config.py
@@ -1,4 +1,6 @@
 import os
+from superset.stats_logger import StatsdStatsLogger
 
 SECRET_KEY = os.environ.get("SECRET_KEY")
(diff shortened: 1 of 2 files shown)


superset/stackable/superset_config.py
```diff
@@ -1,4 +1,6 @@
 import os
+from superset.stats_logger import StatsdStatsLogger
 
 SECRET_KEY = os.environ.get("SECRET_KEY")
 SQLALCHEMY_DATABASE_URI = os.environ.get("SQLALCHEMY_DATABASE_URI")
+STATS_LOGGER = StatsdStatsLogger(host='0.0.0.0', port=9125)
```
maltesander, line 6: Would it make sense to set the metrics port as an env variable from the operator?

maltesander (approved):
LGTM.

PR #52, opened 2022-02-16
Added nginx image.

razvan:
Prerequisite of : https://github.com/stackabletech/regorule-operator/issues/190

### DIFF
--- a/nginx/README.md
+++ b/nginx/README.md
@@ -1,0 +1,18 @@
+# Stackable nginx Docker image
+
+## Description
+
+This image is based on the official (currently 1.21.6) version of the nginx image.
+
+It adds the possibility to upload and download files to `/upload/opa/bundle`. As the name suggests, this is used by the Stackable OPA Operator to upload bundles and by OPA-Services to download them again.
+
+This image stores the files under `/tmp` by default (and for testing purposes), but the OPA operator uses a persistent volume.
+
+## Testing with Docker
+
+  docker run --rm --name nginx -p 8080:80 docker.stackable.tech/stackable/nginx:1.21.6-stackable0
+  echo Upload test.tar.gz ...
+  curl -T test.tar.gz http://localhost:8080/upload/opa/bundle/test.tar.gz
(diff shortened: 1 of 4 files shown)


nginx/README.md
```diff
@@ -0,0 +1,18 @@
...
+
+It adds the possibility to upload and download files to `/upload/opa/bundle`. As the name suggests, this is used by the Stackable OPA Operator to upload bundles and by OPA-Services to download them again.
+
+This image stores the files under `/tmp` by default (and for testing purposes), but the OPA operator uses a persistent volume.
+
+## Testing with Docker
+
+  docker run --rm --name nginx -p 8080:80 docker.stackable.tech/stackable/nginx:1.21.6-stackable0
+  echo Upload test.tar.gz ...
+  curl -T test.tar.gz http://localhost:8080/upload/opa/bundle/test.tar.gz
```
teozkr, line 15: Do we really want/need to allow upload access here? Can't we run the bundle prep stuff as a sidecar instead?

teozkr (changes requested):
This is definitely not suitable for being any kind of default Nginx image. If this goes through (and I'm currently not convinced that it should..) it should at the very least be given a more namespaced name (`opa-regorule-nginx`?).

nginx/README.md
razvan: I simply haven't thought of the sidecar solution. And now that you've mentioned it, I think it is a great idea but I would add that in that case we don't need any service to deliver bundles anymore. OPA can watch a bundle folder and update it's self in place. So a sidecar would just need to update this folder in the OPA pods. What do you think?

razvan: And maybe we don't even need the sidecar if we just mount bundle config maps under the watched bundle root folder.

teozkr: That sounds great. The tricky case for just mounting it would be if we want to load regorules dynamically from multiple configmaps at once.

razvan:
Closing since there seems to be a better solution for opa bundles.

PR #53, opened 2022-02-16
Add python to spark image so that pyspark can be used

adwk67:
Pyspark is already bundled with the spark code that we distribute with this image, but it requires a python installation and also a few core libraries for data wrangling. The changes made here do the following:

- install python and pip
- set the necessary environment variables (e.g. `PYTHONPATH`)
- copies the list of libraries in `requirements.txt` and `pip install`s them (this is done as root, which is advised against where pip libraries are being installed in virtual environments, which is not the case here)

#### Note
- Python libraries have to be pre-installed in the image as a) they should be present on each worker node and b) access to external library repositories is not enabled from the kubernetes pods: this list will expanded as and when specific requirements are identified (e.g. sklearn, h2o, scipy)
- Pyspark jobs can either be tested using the pyspark shell, and then be presented to the cluster as `.py` files and executed with `spark-submit`
- changes to the image were tested with the current spark-operator using example python files that are bundled with the spark code
- pip libraries are not versioned as they are implicitly related to the python version: this may need to be more explicit in the future


spark/requirements.txt
```diff
@@ -0,0 +1,3 @@
+numpy
```
razvan, line 1: It's best practice to pin down the versions of the modules to avoid surprises when building the image sometime in the future.

spark/Dockerfile
```diff
@@ -4,22 +4,38 @@ LABEL maintainer="Stackable GmbH"
...
 COPY spark/stackable /stackable
+COPY spark/requirements.txt /tmp/requirements.txt
+
 RUN groupadd -r stackable --gid=1000 && \
     useradd -r -g stackable --uid=1000 stackable && \
     chown -R stackable:stackable /stackable
 
+# pyspark dependencies
+RUN pip install --upgrade pip
+RUN pip install --no-cache-dir \
```
razvan, line 36: Since spark runs as the `stackable` user, you should be able to use `pip install --user ...` to install under the user's home instead of system wide. But I guess we also need to set the HOME path which is not being currently done.

razvan (commented):
Haven't tried it yet, just some things I saw at first glance.

razvan:
> * pip libraries are not versioned as they are implicitly related to the python version: this may need to be more explicit in the future

This doesn't guarantee that user code stays compatible with eventual upgrades.

adwk67:
> > * pip libraries are not versioned as they are implicitly related to the python version: this may need to be more explicit in the future
> 
> This doesn't guarantee that user code stays compatible with eventual upgrades.

Should we then tie the pip library versions to a spark version: 3.1.1 with numpy a.b.c, 3.2.1 with numpy d.e.f etc.?

razvan:
> Should we then tie the pip library versions to a spark version: 3.1.1 with numpy a.b.c, 3.2.1 with numpy d.e.f etc.? (i.e. have a specific requirements.txt for each spark version)

Sounds good.

razvan (dismissed):
Looks good.

razvan (approved):
lgtm

PR #54, opened 2022-02-17
Add Kafka 3.1.0 image

teozkr:
https://github.com/stackabletech/kafka-operator/issues/309


maltesander (approved):
LGTM. Do we remove the 1.2 authorizer then from the repo?

teozkr:
Personally, I'm not really comfortable deleting artifacts that aren't explicitly in a dev repo, especially when we don't have a clear path path for where the upstream is.

But I think that's a larger issue...

PR #58, opened 2022-02-24
Added the opa authorizer extension

fhennig:
The changes here are quite small, but before this is accepted, the extension itself should be approved as well: https://github.com/stackabletech/druid-opa-authorizer

Once this is merged, a new docker image should be built


fhennig:
Note: What should the version of the extension be? Should it be the same as the druid version, or should the authorizer have its own versioning?

druid/Dockerfile
```diff
@@ -30,7 +30,9 @@ RUN curl -L https://repo.stackable.tech/repository/packages/druid/apache-druid-$
     # Force to overwrite the existing 'run-druid'
     ln -sf /stackable/bin/run-druid /stackable/druid/bin/run-druid && \
     # Install the Prometheus emitter extension. This bundle contains the emitter and all jar dependencies.
-    curl https://repo.stackable.tech/repository/packages/druid/druid-prometheus-emitter-${PRODUCT}.tar.gz | tar -xzC /stackable/druid/extensions
+    curl https://repo.stackable.tech/repository/packages/druid/druid-prometheus-emitter-${PRODUCT}.tar.gz | tar -xzC /stackable/druid/extensions && \
+    # Install OPA authorizer extension.
+    curl https://repo.stackable.tech/repository/packages/druid/druid-opa-authorizer-${PRODUCT}.tar.gz | tar -xzC /stackable/druid/extensions
```
maltesander, line 35: Tagging the authorizer with the druid version (as long as it is not upstream like the emitter) keeps us from fixing anything. Two suggestions:
1) Hardcode the version to something like druid-opa-authorizer-0.1.0 (then we can make a new docker release of druid with a fixed 0.1.1 version)
2) Use a separate entry in the `conf.py` as with Kafka

We dont know when or if we can deploy this upstream, so for now i prefer the flexibility.

maltesander (changes requested):

maltesander (approved):
LGTM.

PR #66, opened 2022-03-11
Nifi monitoring script

maltesander:
- added script to activate nifi monitoring
- added python to image


tools/Dockerfile
```diff
@@ -7,14 +7,17 @@ SHELL ["/bin/bash", "-o", "pipefail", "-c"]
...
+    microdnf install python3-devel python3-pip python3-setuptools && \
     microdnf clean all
 
 # The bcrypt tool is needed by NiFi to locally encrypt the admin password that is mounted as a secret in cleartext
 COPY tools/bin/stackable-bcrypt-1.0-SNAPSHOT-jar-with-dependencies.jar /bin/stackable-bcrypt.jar
+# add all python scripts
+COPY tools/python /stackable/python
 
-RUN groupadd -r stackable --gid=1000 && \
+RUN pip3 install -r /stackable/python/requirements.txt && \
```
soenkeliebau, line 18: Should we not try to keep this in a virtualenv? In case we later add more scripts with different/conflicting dependencies that might make more sense, no?
Then again, virtualenvs in containers must be some sort of antipattern..

tools/python/create_nifi_reporting_task.py
```diff
@@ -0,0 +1,155 @@
...
+
+
+def find_reporting_task(name: str, port: str):
+    """Find a ReportingTask via its name and port"""
+    flow_api = nipyapi.nifi.apis.flow_api.FlowApi()
+
+    try:
+        reporting_tasks = flow_api.get_reporting_tasks().reporting_tasks
+    except Exception as ex:
+        raise Exception("Could not find ReportingTask[{}/{}]: {}"
```
soenkeliebau, line 30: Is this correct? Wouldn't this get thrown if there is an error retrieving the tasks, task could well exist, we just couldn't check..
There should be an error here, just the message is wrong I think. This message should be logged below before returning None..

tools/python/create_nifi_reporting_task.py
```diff
@@ -0,0 +1,155 @@
...
+                        .format(name, port, str(ex))) from None
+
+    for task in reporting_tasks:
```
soenkeliebau, line 33: We get a list back here, right? Couldn't this be done in one line with filter() or something similar?

tools/python/create_nifi_reporting_task.py
```diff
@@ -0,0 +1,155 @@
...
+
+    controller_api = nipyapi.nifi.apis.controller_api.ControllerApi()
+
+    try:
+        return controller_api.create_reporting_task(body=task)
+    except Exception as ex:
+        raise Exception("Could not create reporting task: {}".format(str(ex))) from None
+
+
+def get_reporting_task_id(task):
```
soenkeliebau, line 72: This function seems very brief, do we really need this?

tools/python/create_nifi_reporting_task.py
```diff
@@ -0,0 +1,155 @@
...
+    return task_dict["component"]["state"] == "RUNNING"
+
+
+def set_reporting_task_running(task):
+    """Set ReportingTask to RUNNING"""
+    reporting_task_api = nipyapi.nifi.apis.reporting_tasks_api.ReportingTasksApi()
+
+    state = {
+        "revision": {
+            "version": get_revision_version(task)
```
soenkeliebau, line 101: The revision doesn't change by setting it to running?

tools/python/create_nifi_reporting_task.py
```diff
@@ -0,0 +1,155 @@
...
+
+    reporting_task = find_reporting_task(name=task_name, port=port)
+
+    if reporting_task is None:
+        reporting_task = create_reporting_task(name=task_name, port=port,
+                                               version=args["nifi_version"])
+        print(get_reporting_task_name(task=reporting_task) + " [%s] -> CREATED" % reporting_task.id)
+
+    if not is_reporting_task_running(task=reporting_task):
+        reporting_task = set_reporting_task_running(task=reporting_task)
```
soenkeliebau, line 149: Shouldn't we catch the exceptions thrown by the methods and log something nice?

tools/python/create_nifi_reporting_task.py
```diff
@@ -0,0 +1,155 @@
...
+    # Add arguments to the parser
+    all_args.add_argument("-n", "--nifi_api_url", required=True,
+                          help="The NiFi node url to connect to.")
+    all_args.add_argument("-u", "--username", required=True,
+                          help="Username to connect as.")
+    all_args.add_argument("-p", "--password", required=True,
+                          help="Password for the user.")
+    all_args.add_argument("-v", "--nifi_version", required=True,
+                          help="The NiFi product version.")
+    all_args.add_argument("-c", "--cert", required=True,
```
soenkeliebau, line 127: This could be made a bit clearer .. it is not any public certificate, it is  the certificate authority that we trust and which has to have signed the certificates that we expect..

tools/python/create_nifi_reporting_task.py
```diff
@@ -0,0 +1,155 @@
...
+    task = nipyapi.nifi.models.reporting_task_entity.ReportingTaskEntity(
+        revision=nipyapi.nifi.models.revision_dto.RevisionDTO(version=0),
+        disconnected_node_acknowledged=False,
+        component=nipyapi.nifi.models.reporting_task_dto.ReportingTaskDTO(
+            name=name,
+            type="org.apache.nifi.reporting.prometheus.PrometheusReportingTask",
+            bundle=nipyapi.nifi.models.bundle_dto.BundleDTO(
+                group="org.apache.nifi",
+                artifact="nifi-prometheus-nar",
+                version=version
```
soenkeliebau, line 55: Does this mean that this will break when we update NiFi?

tools/Dockerfile
maltesander: I would make more sense, but i think we can tackle that once we have more scripts.

tools/python/create_nifi_reporting_task.py
maltesander: Youre right it could be a connection problem. Will change.

tools/python/create_nifi_reporting_task.py
maltesander: Yeah im not such a big fan of one liners (pylint isnt either ;)). If i do it in the loop or put a lambda in does not make a big difference to me...

tools/python/create_nifi_reporting_task.py
maltesander: Yeah ill remove and just use task.id.

tools/python/create_nifi_reporting_task.py
maltesander: The exception is the logging here for me.

tools/python/create_nifi_reporting_task.py
maltesander: done

tools/python/create_nifi_reporting_task.py
maltesander: I did not test that. Since it expects a certain version i would assume so. This would mean extending the script to delete or replace outdated ReportingTasks with the newer version.

tools/python/create_nifi_reporting_task.py
soenkeliebau: Ah, that makes sense then, so its to prevent concurrent changes.

tools/python/create_nifi_reporting_task.py
maltesander: The NiFi rolling update does not work. The ReportingTask would be updated if it is stopped properly. If it is running we get into problems with one nifi pod being updated, and trying to update the ReportingTask as well. We need more logic than just the statefulset rolling update for that.

soenkeliebau (approved):

PR #73, opened 2022-03-18
Spark k8s init

adwk67:
Creation of a spark-on-kubernetes image that can be used for driver and executor pods. Changes to the existing spark image:

- use the same entry-point script as spark does internally (`entrypoint.sh`)
- wrap this command with the [tini utility](https://github.com/krallin/tini), as spark does
- pass in the tini utility rather than downloading it
- python elements have been removed as it makes more sense to attach this to job images which will have python dependencies anyway


razvan (approved):
lgtm

Issue #79, opened 2022-03-28
Introduce DNS cache timeouts for Java in all products


lfrancke:
As a user of our operators I'd like to have low Java DNS cache timeouts so the products don't crash on startup.

See these links for reference:
* https://docs.oracle.com/javase/7/docs/technotes/guides/net/properties.html
* https://docs.aws.amazon.com/sdk-for-java/v1/developer-guide/java-dg-jvm-ttl.html

This is done when all Java based products have a `java.security` file (for me it's in `/etc/java/java-11-openjdk/java-11-openjdk-11.0.14.1.1-2.el8_5.x86_64/conf/security` but make sure it's stable) that includes the setting `networkaddress.cache.ttl` and a value of `5`.

For this it might be worth creating one or more (for different Java versions) Stackable Java base images that do this once and we can reuse it for all our images.
See also this for reference: https://github.com/stackabletech/hdfs-operator/issues/147

PR #81, opened 2022-03-28
java-base image for 1.8.0 and 11

maltesander:
Adds a Java base image (currently for 1.8.0 and 11) to be used in the operator images depending on java.

precondition for https://github.com/stackabletech/docker-images/issues/79

### DIFF
--- a/java-base/Dockerfile
+++ b/java-base/Dockerfile
@@ -1,4 +1,4 @@
-FROM registry.access.redhat.com/ubi8/ubi-minimal
+FROM registry.access.redhat.com/ubi8/ubi-minimal:8.5
 LABEL maintainer="Stackable GmbH"
 
 # https://github.com/hadolint/hadolint/wiki/DL4006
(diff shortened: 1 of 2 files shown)


java-base/Dockerfile
```diff
@@ -0,0 +1,16 @@
...
+LABEL maintainer="Stackable GmbH"
+
+# https://github.com/hadolint/hadolint/wiki/DL4006
+SHELL ["/bin/bash", "-o", "pipefail", "-c"]
+
+ARG PRODUCT
+ARG _SECURITY_PATH
+
+RUN microdnf update && \
+    microdnf install java-${PRODUCT}-openjdk-headless --nodocs && \
```
sbernauer, line 11: Just wanted to make sure if we want to install a JRE or a JDK?
From the Image size and security perspective i would think a JRE like `java-${PRODUCT}-openjdk-jre-headless` should be sufficient

sbernauer (changes requested):
LGTM besides comment and failing check

java-base/Dockerfile
sbernauer: From the daily JDK is fine for now as some products need a JDK

sbernauer (dismissed):

sbernauer (approved):

PR #83, opened 2022-03-29
Using java-base image in java based tools

maltesander:
- switched to java-base image (11 for everything but druid)
- removed dnf java install 
- removed java home env var

closes https://github.com/stackabletech/docker-images/issues/79


sbernauer:
Just a small note: We are now changing the JAVA_HOME from `ENV JAVA_HOME /etc/alternatives/jre` to `ENV JAVA_HOME=/usr/lib/jvm/jre-${PRODUCT}`
Fine for me, just wanted to point that out

maltesander:
I think we had `JAVA_HOME=/usr/lib/jvm/jre-${PRODUCT}` everywhere except for in druid? What would be the best practice? I thin pointing to the alternatives would be much cleaner?

soenkeliebau:
I think alternatives is useful more in a system that actually runs separate workloads that might need separate java versions. In our case in a container I think we can safely use the specific link to the exact version to make things more specific.

PR #97, opened 2022-04-22
Rebase airflow image on Redhat UBI

### LINKED ISSUE #89: Base Airflow image on the Red Hat UBI
All images must fulfill the requirements for OpenShift (see stackabletech/issues#207). One requirement is that all images must be based on a Red Hat Enterprise Linux or Red Hat Universal Base Image (see https://redhat-connect.gitbook.io/partner-guide-for-red-hat-openshift-and-container/program-on-boarding/technical-prerequisites#dockerfile-requirements). The Stackable Airflow image is based on the upstream Airflow image which in the end is based on Debian. Therefore the base image must be replaced with a Red Hat UBI.

The upstream Airflow image contains an argument `PYTHON_BASE_IMAGE` (see https://github.com/apache/airflow/blob/2.2.5/Dockerfile#L47). It should be evaluated if this can be set to an image which is based on a Red Hat UBI.

adwk67:
Replace Debian-based Airflow image with one based on Redhat UBI.
Closes #89.


airflow/CHANGELOG.md
```diff
@@ -1,5 +1,17 @@
 # Changelog
 
+## [Unreleased]
+
+### Added
+
+- Added Redhat-UBI-based image ([#97]).
```
teozkr, line 7: Isn't this more of a changed entry?

adwk67: This change will result in images that are named differently (the python variable is used for dnf installs rather than locating the airflow image), so I will delete the old ones when the build action has been done. But, yes, it is essentially a change: I'll edit that.

airflow/Dockerfile
```diff
@@ -20,3 +13,52 @@ LABEL name="Apache Airflow" \
...
 COPY airflow/licenses /licenses
+COPY airflow/constraints-${PRODUCT}-python${PYTHON}.txt /tmp/constraints.txt
+
+# the mysql provider is currently excluded.
+# Requires implementation of https://github.com/apache/airflow/blob/2.2.5/scripts/docker/install_mysql.sh
+ENV AIRFLOW_EXTRAS=async,amazon,celery,cncf.kubernetes,docker,dask,elasticsearch,ftp,grpc,hashicorp,http,ldap,google,google_auth,microsoft.azure,odbc,pandas,postgres,redis,sendgrid,sftp,slack,ssh,statsd,virtualenv
+
+RUN microdnf update \
+        && microdnf install -y \
+            gcc \
```
razvan, line 24: The dependencies required to build libraries should not be included in the final image. We should use a multi-staged Docker image similar to how we do it for operator images with the rust toolchain. (https://docs.docker.com/develop/develop-images/multistage-build/)

razvan: To make the transfer of compiled libs/packages/etc to the final stage a bit less painful, we should probably use a virtual env or `--user` when calling `pip install`.

adwk67: The image now uses a multi-stage build with the virtual-env being copied across.

airflow/Dockerfile
```diff
@@ -1,15 +1,46 @@
...
+
+ENV HOME=/stackable
+ENV AIRFLOW_USER_HOME_DIR=/stackable
+ENV PATH=$PATH:/bin:$HOME/app/bin
+ENV AIRFLOW_HOME=$HOME/airflow
+
+RUN mkdir -pv ${HOME} && \
+    mkdir -pv ${AIRFLOW_HOME} && \
+    mkdir -pv ${AIRFLOW_HOME}/dags && \
+    mkdir -pv ${AIRFLOW_HOME}/logs
```
razvan, line 37: I think this entire block has no effect since you start from a fresh image right after it and you have to duplicate it below.

airflow/Dockerfile
```diff
@@ -20,3 +51,38 @@ LABEL name="Apache Airflow" \
       description="This image is deployed by the Stackable Operator for Apache Airflow."
 
 COPY airflow/licenses /licenses
+
+RUN microdnf update \
+        && microdnf install -y \
+            shadow-utils \
+            python${PYTHON} \
+        && microdnf clean all
```
razvan, line 59: This suggestion should have the same effect and make the image better for OpenShift certification. We have to update all images sometime in the future with this block.
```suggestion
# Update image and install python
RUN microdnf install -y yum python${PYTHON} \
  && yum -y update-minimal --security --sec-severity=Important --sec-severity=Critical \
  && yum clean all \
  && microdnf clean all 
```

airflow/Dockerfile
```diff
@@ -1,15 +1,46 @@
...
+        && microdnf install -y \
+            gcc \
+            gcc-c++ \
+            python${PYTHON}-devel \
+            python${PYTHON}-pip \
+            python${PYTHON}-wheel \
+            openssl-devel \
+            cyrus-sasl-devel \
+            openldap-devel \
+            unixODBC-devel \
```
razvan, line 22: I think you have to install non devel versions in the final stage for any code that uses them to work.

airflow/Dockerfile
adwk67: Airflow is installed under `/stackable` so it's part of the structure that gets copied over to the new image.

adwk67: Changed.

airflow/Dockerfile
```diff
@@ -1,15 +1,46 @@
...
+RUN mkdir -pv ${HOME} && \
+    mkdir -pv ${AIRFLOW_HOME} && \
+    mkdir -pv ${AIRFLOW_HOME}/dags && \
+    mkdir -pv ${AIRFLOW_HOME}/logs
+
+FROM registry.access.redhat.com/ubi8/ubi-minimal:8.5 as airflow-main-image
+
+ARG PRODUCT
+ARG PYTHON
+ARG RELEASE="1"
```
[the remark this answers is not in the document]
adwk67: This does not appear to work.

airflow/Dockerfile
adwk67: Done. Will rebuild against supported versions and re-run the kuttl tests.

airflow/Dockerfile
adwk67: Done.

razvan (approved):
lgtm

PR #103, opened 2022-04-28
Base Superset image on the Red Hat UBI

### LINKED ISSUE #88: Base Superset image on the Red Hat UBI
All images must fulfill the requirements for OpenShift (see stackabletech/issues#207). One requirement is that all images must be based on a Red Hat Enterprise Linux or Red Hat Universal Base Image (see https://redhat-connect.gitbook.io/partner-guide-for-red-hat-openshift-and-container/program-on-boarding/technical-prerequisites#dockerfile-requirements). The Stackable Superset image is based on the upstream Superset image which in the end is based on Debian. Therefore the base image must be replaced with a Red Hat UBI.

siegfriedweber:
Closes #88

### DIFF
--- a/conf.py
+++ b/conf.py
@@ -118,11 +118,11 @@ products = [
         'versions': [
             {
                 'product': '1.3.2',
-                '_base_image_tag': '286ba5d37d4e240d01bbefd2307a816829cf512d',
+                'python': '38',
             },
             {
                 'product': '1.4.1',
-                '_base_image_tag': '26545a26d195b79ab6e838631e86cb0bf20f3ced',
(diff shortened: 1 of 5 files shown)


adwk67 (commented):
A minor point (relating to usage rather than the image build itself): the image names now have a python element to them so the example in the operator should have a corresponding change e.g. `1.4.1-python39` rather than just `1.4.1`. Likewise with the cluster names in the integration tests.

adwk67 (approved):
Re. comment about image names - I missed the fact that the build script will take care of this!
LGTM.

PR #108, opened 2022-05-03
Set version label in Superset image

siegfriedweber:
Set version label in Superset image

The version label was empty due to a missing `ARG` instruction in the Dockerfile.


maltesander (approved):
LGTM.

PR #110, opened 2022-05-04
Add Python package pydruid to Superset image

siegfriedweber:
Add Python package pydruid to Superset image

This package contains a database driver which is necessary for Superset to connect to Apache Druid.

### DIFF
--- a/superset/Dockerfile
+++ b/superset/Dockerfile
@@ -35,6 +35,7 @@ RUN microdnf update \
         gevent \
         psycopg2-binary \
         statsd \
+        pydruid \
         python-ldap
 
 # Final image
--- a/superset/CHANGELOG.md
+++ b/superset/CHANGELOG.md
@@ -1,5 +1,13 @@
 # Changelog
 
+## [superset-stackable2.1.0] - 2022-05-04
+
+### Added
+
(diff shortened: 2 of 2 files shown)


fhennig (approved):

superset/Dockerfile
```diff
@@ -35,6 +35,7 @@ RUN microdnf update \
         gevent \
         psycopg2-binary \
         statsd \
+        pydruid \
```
sbernauer, line 38: Just a hint: I couldn't find `pydruid` in the constraints.txts. So it seems to me we don't install a fixed version

sbernauer (approved):

sbernauer:
damn, too slow

superset/Dockerfile
siegfriedweber: The constraints contain only the libraries which are non-optional dependencies of Superset. This is necessary because otherwise Superset cannot be built or run. All additional libraries are not contained. If adding these libraries then it would make sense to add all transitive dependencies, too.

As the system packages are upgraded to contain the latest bug fixes, I thought it would be a good idea to not constrain the Python packages either if it is not necessary.

PR #114, opened 2022-05-10
Call "superset init" before starting Superset

sbernauer:
We need this for Superset LDAP integration
See https://github.com/stackabletech/superset-operator/pull/180#pullrequestreview-965673428

### DIFF
--- a/superset/CHANGELOG.md
+++ b/superset/CHANGELOG.md
@@ -5,6 +5,7 @@
 ### Added
 
 - Python package pydruid added ([#110]).
+- Call `superset init` before starting Superset ([#114](https://github.com/stackabletech/docker-images/pull/114))
 
 [#110]: https://github.com/stackabletech/docker-images/pull/110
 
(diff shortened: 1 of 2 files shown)


superset/CHANGELOG.md
```diff
@@ -5,6 +5,7 @@
 ### Added
 
 - Python package pydruid added ([#110]).
+- Call `superset init` before starting Superset ([#114](https://github.com/stackabletech/docker-images/pull/114))
```
siegfriedweber, line 8: This is not part of `superset-stackable2.1.0` and must be moved to a new version. The other entries do not reference the URL inline.

superset/Dockerfile
```diff
@@ -83,7 +83,8 @@ USER stackable
 WORKDIR ${HOME}
 
 CMD ["/bin/sh", "-c", \
-    "gunicorn \
+    "superset init && \
```
siegfriedweber, line 86: This is a possible solution. Another (more elegant) solution would be to add this to the pod command or create an init container.

siegfriedweber (changes requested):

PR #116, opened 2022-05-12
Add patch to fix Flask LDAP bug in Superset

sbernauer:
Apply the patch we PRed at Flask-Appbuilder: https://github.com/dpgaspar/Flask-AppBuilder/pull/1846


lfrancke (approved):
when/if we have more patches we might need a more structured approach but for now I'm fine with this

sbernauer:
Totally agree

PR #119, opened 2022-05-20
Updated log4shell_scanner with support for aarch64 builds

Jimvin:
I've added the arm64 version of log4shell_scanner and a wrapper script to run the correct version depending on the container architecture. The Java based images now build successfully on ARM.

### DIFF
--- a/shared/log4shell_scanner
+++ b/shared/log4shell_scanner
@@ -1,3 +1,3 @@
 #!/bin/sh
-ARCH="$(arch)"
+ARCH="$(uname --machine)"
 /bin/log4shell_scanner_${ARCH} "$@"
--- a/shared/log4shell_scanner
+++ b/shared/log4shell_scanner
@@ -1,3 +1,3 @@
-#!/bin/bash
-ARCH=`arch`
-/bin/log4shell_scanner_${ARCH} $*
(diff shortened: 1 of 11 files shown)


shared/log4shell_scanner
```diff
@@ -0,0 +1,3 @@
+#!/bin/bash
+ARCH=`arch`
+/bin/log4shell_scanner_${ARCH} $*
```
teozkr, line 3:
```suggestion
#!/bin/sh
ARCH="$(arch)"
/bin/log4shell_scanner_${ARCH} "$@"
```

teozkr: `arch` is a weird one, it seems to exist in our UBI base image and is documented as part of coreutils, but it doesn't seem to be packaged at all on Arch (although `man uname` refers to it?!).

Would it make sense to use `uname --machine` instead?

teozkr (approved):

shared/log4shell_scanner
Jimvin: Sensible suggestions. Yes, uname --machine is likely to have wider support over different Linux distros.

teozkr (approved):

PR #122, opened 2022-05-26
Fixed tools image to build on arm64 architecture.

Jimvin:


teozkr:
Looks good to me, but do we have a good way to test this on x86 machines?

teozkr (approved):
This seems to build and run fine for me under QEMU, but if we want to consider this a supported configuration then we'll probably need to figure out a less painful way to test it at some point.

Jimvin:
I've run a test build in my x86 workstation and all of the images build successfully. Ideally we need to build and test the images for any supported architectures.

PR #124, opened 2022-05-30
Remove dependency versions from Docker image tags

teozkr:
This is tracked as part of the image/platform version anyway, and was a weird detail to require users to configure manually.

We could keep building the dep-tagged images as well as a fallback, but this doesn't seem worthwhile when they were never used anyway.

This applies to many of the ADR018 PRs already created, such as https://github.com/stackabletech/druid-operator/pull/238 and https://github.com/stackabletech/airflow-operator/pull/106.

### DIFF
--- a/build_product_images.py
+++ b/build_product_images.py
@@ -52,38 +52,19 @@ def build_image_tags(image_name, image_version, product_version):
     """
     Returns a list of --tag command line arguments that are used by the
     docker build command.
-    Each image is tagged with three tags as follows:
-        1. <product>-<dependency1>-<dependency2>...-<image>
-        2. <product>-<dependency1>-<dependency2>...-<platform>
-        3. <product>-<platform>
-
-    Product version items starting with an underscore are not appended as
-    dependencies.
+    Each image is tagged with two tags as follows:
+        1. <product>-<image>
+        2. <product>-<platform>
     """
-    result = []
 
     platform_version = re.search(r'^\d+', image_version)[0]
-
     if isinstance(product_version, dict):
-        dep_versions = [f'{key}{value}' for key, value in product_version.items() if key != "product" and not key.startswith('_')]
-        image_tag = "-".join([product_version['product'], *dep_versions, f'stackable{image_version}'])
-        platform_tag = "-".join([product_version['product'], *dep_versions, f'stackable{platform_version}'])
-        latest_tag = "-".join([product_version['product'], f'stackable{platform_version}'])
+        product_version = product_version['product']
 
-        result.extend([
-            '-t', f'{image_name}:{image_tag}',
-            '-t', f'{image_name}:{platform_tag}',
-            '-t', f'{image_name}:{latest_tag}'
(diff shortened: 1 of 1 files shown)


soenkeliebau:
Just to make sure I understand this correctly, this means that we'd bake for example the authorizer version into the stackable image version?
So for example the following transformations would apply:
1.2.3-authorizer0.1-stackable0.1.0 -> 1.2.3-stackable0.1.0
1.2.3-authorizer0.2-stackable0.0.1 -> 1.2.3-stackable0.2.0

and subesequent updates to the Dockerfiles (if we even want to support older authorizer versions) would need to be backported to the 0.1.0 branch?
So if we update the baseimage for example, we'd release:
1.2.3-stackable0.1.1 (containing authorizer 0.1)
1.2.3-stackable0.2.1 (containing authorizer 0.2)

Did I get that roughly correct?

teozkr:
Yes, just like for the rest of the environment that we bake implicitly into the stackable platform release version. The three-dimensional support matrix of (operator, product, platform) is confusing (and slow to build) enough on its own.

teozkr:
Also not sure if this was intentional, but your example jumps from authorizer 0.1.0 with platform 0.1.0 to authorizer 0.2.0 with platform 0.0.1 (lower than the 0.1.0 before). I don't really see a valid case for that to happen, so I'm assuming that that was a mistake.

soenkeliebau:
> Also not sure if this was intentional, but your example jumps from authorizer 0.1.0 with image 0.1.0 to authorizer 0.2.0 with image 0.0.1 (lower than the 0.1.0 before). I don't really see a valid case for that to happen (at least not without also providing a image 0.1.0 build of authorizer 0.2.0 and vice versa), so I'm assuming that that was a mistake.

 I have corrected the mistake that I saw in my example - not 100% sure it is the one you saw tbh :)

adwk67:
The only observation I'd add is that this makes the accuracy of the changelog essential, as that should indicate what was added and when. We will probably end up having multiple dependencies (say, if we add statsd as a dependency to the airflow and superset images) yielding something like:

```
1.2.3-dep1-stackable0.1.0          -> 1.2.3-stackable0.1.0
1.2.3-dep1-dep2-stackable0.1.0     -> 1.2.3-stackable0.2.0
```
and the information about where a backport to a particular version of dep1 belongs should be something that is clear from the changelog (which is what we intend).

teozkr:
Yes.

adwk67:
This means that we cannot release different versions of, say, python for a given version of airflow. I guess that's not a problem, as  we could always build a specific dep-tagged image if anyone needs one, and it's probably better to keep things streamlined until such time as there is a clear demand for that.

soenkeliebau:
If the two of you are happy then so am I :)

Not 100% grasped the discussion I think, but I can fully get behind going with this until we have a specific need for.aomething more complex.

soenkeliebau (approved):

PR #156, opened 2022-07-22
Cross-platform building

Maleware:
FIrst draft for cross-plattfrom compiling.

There are some braking changes for our github action. Giving the script -c will bring it back to normal but can be used with -a/--architecture arm64/amd64/... .  For our actions we can use the option -a to cross-compile. Be aware that not giving -a will result in --platform platform.machine(), so the native platform will be passed to docker. 

**Actual Issue:**
1. Java-Base and Tools can not build with --platform flag, we run into an error from redhead, feel free to try and comment your ideas. (**Solved**)
2. You can not give multiple platforms ant once like -a amd64, arm64 since buildx is not supporting multi-platform. That would be a next step if my approach is decent for now.

### DIFF
--- a/build_product_images.py
+++ b/build_product_images.py
@@ -49,7 +49,7 @@ def parse_args():
     )
     parser.add_argument("-u", "--push", help="Push images", action="store_true")
     parser.add_argument("-d", "--dry", help="Dry run.", action="store_true")
-    parser.add_argument("-a", "--architecture", help="Target platform for image")
+    parser.add_argument("-a", "--architecture", help="Target platform for image", nargs='*')
     parser.add_argument("-m", "--multiarch", help="Build and publish multi-architecture images", action="store_true")
     parser.add_argument("-n", "--node", help="Create nodes to a builder. First is Master.", type=str)
     return parser.parse_args()
@@ -112,7 +112,7 @@ def build_and_publish_image(args, product):
         push = "--push"
 
     # Multiarch builds
-    if args.multiarch:
+    if len(args.architecture) > 1:
         commands.append(
             [
                 "docker",
@@ -123,7 +123,7 @@ def build_and_publish_image(args, product):
                 "-f",
                 product["name"] + "/Dockerfile",
                 "--platform",
-                args.architecture,
+                ','.join(args.architecture),
                 push,
                 ".",
             ]
@@ -140,7 +140,7 @@ def build_and_publish_image(args, product):
                 "-f",
                 product["name"] + "/Dockerfile",
                 "--platform",
-                check_platform(args.architecture),
+                ','.join(args.architecture),
                 "--load",
                 ".",
             ]
@@ -197,7 +197,7 @@ def check_platform(architecture):
     Checks if a desired platform is given, gives current platform if not
     """
     if architecture is None:
-        architecture = "linux/" + platform.machine()
+        architecture = ["linux/" + platform.machine()]
 
     return architecture
 
@@ -283,7 +283,9 @@ def main():
     args = parse_args()
     print("Current Platform: ", platform.machine())
 
-    if args.multiarch:
+    args.architecture = check_platform(args.architecture)    
+
+    if len(args.architecture) > 1:
         if args.node is None:
             create_virtual_enviroment(args)
 
@@ -302,7 +304,7 @@ def main():
 
     run_commands(args.dry, commands)
 
-    if args.multiarch:
+    if len(args.architecture) > 1:
         if args.node is None:
             remove_virtual_enviroment(args)
 
--- a/build_product_images.py
+++ b/build_product_images.py
@@ -251,7 +251,7 @@ def main():
     run_commands(args.dry, commands)
 
     if len(args.architecture) > 1:
-            remove_virtual_enviroment(args)
+        remove_virtual_enviroment(args)
 
 
 if __name__ == "__main__":
--- a/build_product_images.py
+++ b/build_product_images.py
@@ -218,7 +218,7 @@ def remove_virtual_enviroment(args):
 
 def check_architecture_input(args):
 
-    supported_arch = ["linux/amd64","linux/arm64"]
+    supported_arch = ["linux/amd64", "linux/arm64"]
 
     for target_arch in args.architecture:
         if target_arch not in supported_arch:
--- a/build_product_images.py
+++ b/build_product_images.py
@@ -191,7 +191,7 @@ def check_architecture_input(architecture):
 
     if architecture not in supported_arch:
         raise ValueError(
-            f"Architecture {architectures} not supported. Supported: {supported_arch}"
+            f"Architecture {architecture} not supported. Supported: {supported_arch}"
         )
     
     return architecture
--- a/build_product_images.py
+++ b/build_product_images.py
@@ -10,15 +10,12 @@ Example:
     build_product_images.py --product zookeeper,kafka --image_version 0.1.0 --push
 
 This will build an image for each Apache ZooKeeper and Apache Kafka version configured in conf.py
-"""
 
-"""
-as a reminder:
-If we want to build parallel we have to use building nodes like the following example suggests (from https://docs.docker.com/build/buildx/):
-docker buildx create --use --name mybuild node-amd64
-docker buildx create --append --name mybuilde node-arm64
-
-docker buildx build --platform linux/amd64, linux/arm64
+Multiarchitecture builds:
+This assumes that the following images are available for target architecture:
+    1. Java-Base 11, 1.8.0
+    2. ubi8-rust-builder
+    3. Tools 0.2.0
 
 """
 
@@ -53,7 +50,7 @@ def parse_args():
     parser.add_argument("-d", "--dry", help="Dry run.", action="store_true")
     parser.add_argument("-a", "--architecture", help="Target platform for image")
     parser.add_argument("-c", "--check", help="Setting the flag will enable dependency checks and building layers", action="store_true")
-    parser.add_argument("-m", "--multiarch", help="Build and publish multi-arch images", action="store_true")
+    parser.add_argument("-m", "--multiarch", help="Build and publish multi-architecture images", action="store_true")
     return parser.parse_args()
 
 
@@ -95,12 +92,13 @@ def build_image_tags(image_name, image_version, product_version):
         f"{image_name}:{product_version}-stackable{image_version}",
     ]
 
-# Not sure if --load is always required. Testing revealed that it won't hurt, in case we want to use a builder instants evoked 
-# by docker buildx create --use, --load is required to bring the image out of the container
+
 def build_and_publish_image(args, product):
     """
     Returns a list of commands that need to be run in order to build and
     publish product images.
+
+    For local building, builder instances are supported.
     """
     commands = []
 
@@ -188,7 +186,10 @@ def product_to_build(product_name, product_version, products):
     else:
         return None
 
-############################################################### Local dependency checks ####################################################################
+"""
+In generall, the following checks weather or not dependencies are available on your local machine. This is only useful if a local repository via Docker-Desktop or
+something simular is used. However, this is not checking if a dependency of certain architecutre is existent in a repository!
+"""
 
 def check_or_build_dependencies(args, architecture, products):
     """
@@ -222,7 +223,6 @@ def build_dependencies(java, tools, rust_builder, args, products):
     Builds neccessary dependencies for images if not available on system
     """
 
-    # edenhall/kcat is not checked yet, since it is supposed to be moved to tools. But not sure about that.
     args_dummy = copy.deepcopy(args)
 
     if not rust_builder:
@@ -234,20 +234,20 @@ def build_dependencies(java, tools, rust_builder, args, products):
         args_dummy.product_version = '11'
         print('Building dependency Java-Base', args_dummy.product_version)
 
-        run_commands(args_dummy.dry, build_and_publish_image(args_dummy, product_to_build('java-base', '11', products)))
+        run_commands(args_dummy.dry, build_and_publish_image(args_dummy, product_to_build('java-base', args_dummy.product_version, products)))
 
         args_dummy.image_version = '0'
         args_dummy.product_version = '1.8.0'
         print('Building dependency Java-Base', args_dummy.product_version)
 
-        run_commands(args_dummy.dry, build_and_publish_image(args_dummy, product_to_build('java-base', '1.8.0', products)))
+        run_commands(args_dummy.dry, build_and_publish_image(args_dummy, product_to_build('java-base', args_dummy.product_version, products)))
 
     if not tools:
         args_dummy.image_version = '0'
         args_dummy.product_version = '0.2.0'
         print("Building dependency Tools", args_dummy.product_version)
 
-        run_commands(args_dummy.dry, build_and_publish_image(args_dummy, product_to_build('tools', '0.2.0', products)))
+        run_commands(args_dummy.dry, build_and_publish_image(args_dummy, product_to_build('tools', args_dummy.product_version, products)))
 
 
 def check_platform(architecture):
@@ -259,7 +259,7 @@ def check_platform(architecture):
 
     return architecture
 
-############################################################### Multi-Arch-Images #########################################################################
+
 def create_virtual_enviroment(args):
    
     commands=[]
--- a/build_product_images.py
+++ b/build_product_images.py
@@ -16,7 +16,6 @@ This assumes that the following images are available for target architecture:
     1. Java-Base 11, 1.8.0
     2. ubi8-rust-builder
     3. Tools 0.2.0
-
 """
 
 import conf
@@ -108,7 +107,7 @@ def build_and_publish_image(args, product):
 
     # Multiarch builds
     if args.multiarch:
-       commands.append(
+        commands.append(
             [
                 "docker",
                 "buildx",
@@ -118,12 +117,12 @@ def build_and_publish_image(args, product):
                 "-f",
                 product["name"] + "/Dockerfile",
                 "--platform",
-                "linux/amd64,linux/arm64", 
+                "linux/amd64,linux/arm64",
                 "--push",
                 ".",
             ]
         )
-    #local builds / single architecture
+    # local builds / single architecture
     else:
         commands.append(
             [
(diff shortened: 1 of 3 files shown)


build_product_images.py
```diff
@@ -149,8 +192,164 @@ def product_to_build(product_name, product_version, products):
...
+
+    run_commands(args.dry, commands)
+
+
 def main():
     args = parse_args()
+    print("Current Platform: ", platform.machine())
+
+    if args.check:
+        check_or_build_dependencies(args, check_platform(args.architecture), conf.products)
```
razvan, line 345: I would remove this for multiple reasons:
- the versions of the requirements are hard coded in this script.
- it introduces an additional dependency `import docker` which needs to be installed everywhere.
- this script is not a dependency manager

build_product_images.py
```diff
@@ -85,26 +99,55 @@ def build_and_publish_image(args, product):
...
+                "docker",
+                "buildx",
+                "build",
+                *build_args,
+                *tags,
+                "-f",
+                product["name"] + "/Dockerfile",
+                "--platform",
+                "linux/amd64,linux/arm64",
+                push,
```
razvan, line 127: Is this `push` needed here? The actual pushing of the images is done as an additional (optional) step below. This way I can always build the images locally without pushing them. Needed for testing.

build_product_images.py
```diff
@@ -39,6 +49,10 @@ def parse_args():
     )
     parser.add_argument("-u", "--push", help="Push images", action="store_true")
     parser.add_argument("-d", "--dry", help="Dry run.", action="store_true")
+    parser.add_argument("-a", "--architecture", help="Target platform for image")
+    parser.add_argument("-c", "--check", help="Setting the flag will enable dependency checks and building layers", action="store_true")
+    parser.add_argument("-m", "--multiarch", help="Build and publish multi-architecture images", action="store_true")
```
razvan, line 54: This is not needed. I can pass multiple architectures with the `-a` argument.

build_product_images.py
```diff
@@ -85,26 +99,55 @@ def build_and_publish_image(args, product):
...
+            [
+                "docker",
+                "buildx",
+                "build",
+                *build_args,
+                *tags,
+                "-f",
+                product["name"] + "/Dockerfile",
+                "--platform",
+                "linux/amd64,linux/arm64",
```
razvan, line 126: Just use the contents of `args.architecture` here.

build_product_images.py
```diff
@@ -149,8 +192,164 @@ def product_to_build(product_name, product_version, products):
...
+
+    if args.check:
+        check_or_build_dependencies(args, check_platform(args.architecture), conf.products)
+
+    if args.multiarch:
+        if args.node is None:
+            create_virtual_enviroment(args)
+
+    if args.node is not None:
+        join_nodes(args)
```
razvan, line 352: Again, I never used `buildx` ... is this some form of speedup for local builds ? Because it's not used in the GH action. Is it really needed?

razvan (commented):
This needs work. Here is a summary of things I think need to be changed:
- Remove entire the dependency building stuff and the import docker
- The` -m` and `-a` arguments are inconsistent which each other. IMO `-m` is not needed.
- The dry run (`-d`) is now broken because` run_commands() `is called all over the place now. The idea was to collect all commands and run them in a single step at the end of `main`. I would like to preserve this behavior.

build_product_images.py
Maleware: This because if you build multi arch and give docker --platform with at least two target platforms, it needs a different builder to be happy. Currently we do not have native worker for arm64, therefore in the beginning the emulated for arm64 will be used. 
Due to poor performance, I suspect us using native nodes at some point. That's meant to be the preparation for building multi arch with native nodes.

build_product_images.py
Maleware: It seems to be. When I've tried to separate the comments it happens when you want at least two target platforms, the images without --push won't be loaded to registry, therefore with docker push --all-tags the multi image can not be pushed to the repository. 

To work around that, if -u is not set the variable push will contain an empty string. If -u is given, push contains "--push" and the image will be send to the repository.

build_product_images.py
Maleware: Done

build_product_images.py
Maleware: Done

build_product_images.py
Maleware: Done

build_product_images.py
Maleware: Done, since we do not use native worker right now.

build_product_images.py
```diff
@@ -149,8 +185,55 @@ def product_to_build(product_name, product_version, products):
         return None
 
 
+def check_platform(architecture):
+    """
+    Checks if a desired platform is given, gives current platform if not
+    """
+    if architecture is None or len(architecture) == 0:
```
razvan, line 192: This is enough:
```suggestion
    if not architecture:
```

build_product_images.py
```diff
@@ -85,26 +92,55 @@ def build_and_publish_image(args, product):
...
-            product["name"] + "/Dockerfile",
-            ".",
-        ]
-    )
     if args.push:
-        commands.append(["docker", "push", "--all-tags", image_name])
+        push = "--push"
+
+    # Multiarch builds
+    if len(args.architecture) > 1:
```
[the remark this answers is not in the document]
Maleware: Would be fine by me, downside would be that if a builder is up and running because somebody did `docker buildx create --use`, without `--load` the image will not be loaded into the registry. On a fresh environment, that would never be the case I guess and docker will give a hint: `WARNING: No output specified for docker-container driver. Build result will only remain in the build cache. To push result image into registry use --push or to load image into docker use --load` .

This might be hint enough, if we do not care about this I am happy to remove it :)

build_product_images.py
```diff
@@ -149,8 +185,55 @@ def product_to_build(product_name, product_version, products):
         return None
 
 
+def check_platform(architecture):
+    """
+    Checks if a desired platform is given, gives current platform if not
+    """
+    if architecture is None or len(architecture) == 0:
+        architecture = ["linux/" + platform.machine()]
```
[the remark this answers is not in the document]
Maleware: I don't know, I tested now loosely what would happen if I give docker a --platform linux/X86_64 and it looks like it still fetches an amd64. I tested it with tools and with java-base 1.8.0, inspecting those images suggest that it is amd64.

If you think this might end in odd behaviour or errors which could be avoided I am still happy to delete that too.

build_product_images.py
Maleware: I agree, done

build_product_images.py
```diff
@@ -45,7 +45,7 @@ def parse_args():
     )
     parser.add_argument("-u", "--push", help="Push images", action="store_true")
     parser.add_argument("-d", "--dry", help="Dry run.", action="store_true")
-    parser.add_argument("-a", "--architecture", help="Target platform for image, Expecting -a <platform 1> <platform 2> ...", nargs='+')
+    parser.add_argument("-a", "--architecture", help="Target platform for image, Expecting -a <platform 1> <platform 2> ... At least one argument", nargs='+', required=True)
```
razvan, line 48: I would add a little validation as here as described here: https://stackoverflow.com/a/37472037

We also have an example in the (now defunct) integration-tests repo: https://github.com/stackabletech/integration-tests/blob/8114543fc51b7cd66cec50435cb791b274babec7/create_test_cluster.py#L116

build_product_images.py
```diff
@@ -10,13 +10,19 @@
     build_product_images.py --product zookeeper,kafka --image_version 0.1.0 --push
```
adwk67, line 10: This example was already out-of-date before this PR (as product-version is omitted): we could either drop it (as the information is there with the -h option) or add a couple of examples to reflect the newest changes.

build_product_images.py
```diff
@@ -39,6 +45,7 @@ def parse_args():
     )
     parser.add_argument("-u", "--push", help="Push images", action="store_true")
     parser.add_argument("-d", "--dry", help="Dry run.", action="store_true")
+    parser.add_argument("-a", "--architecture", help="Target platform for image, Expecting -a <platform 1> <platform 2> ... At least one argument", nargs='+', required=True)
```
adwk67, line 48: `buildx` doesn't work out of the box for users not using docker-desktop and has to be installed separately. Maybe just add a line in the description to make that clear, so it reads like e.g.: "Build and publish product images. See conf.py for details regarding product versions.You may need to install buildx as a cli-plugin if not using docker-desktop: see the Docker site for details."

razvan: Done.

build_product_images.py
razvan: Updated.

razvan (dismissed):

build_product_images.py
```diff
@@ -81,32 +101,41 @@ def build_image_tags(image_name, image_version, product_version):
     ]
 
 
-def build_and_publish_image(args, product):
+def build_and_publish_image(args, product) -> list[list[str]]:
```
sbernauer, line 104: Python 3.8.10 returns
```
python3 ./build_product_images.py 
Traceback (most recent call last):
  File "./build_product_images.py", line 104, in <module>
    def build_and_publish_image(args, product) -> list[list[str]]:
TypeError: 'type' object is not subscriptable
```
Use together with `from typing import List`
```suggestion
def build_and_publish_image(args, product) -> List[List[str]]:
```

build_product_images.py
```diff
@@ -1,35 +1,47 @@
 #!/usr/bin/env python
 """
-Build and possibly publish product images. It doesn't login to any registry when publishing
-but it assumes a `docker login` has been performed before.
+Build and possibly publish product images.
```
sbernauer, line 3: Did a quick `pylint build_product_images.py`. You can work on those if you want to but you don't have to.
```
build_product_images.py:77:15: C0103: Variable name "v" doesn't conform to snake_case naming style (invalid-name)
build_product_images.py:104:46: E1136: Value 'list' is unsubscriptable (unsubscriptable-object)
build_product_images.py:104:51: E1136: Value 'list' is unsubscriptable (unsubscriptable-object)
build_product_images.py:148:12: W1510: Using subprocess.run without explicitly set `check` is not recommended. (subprocess-run-check)
build_product_images.py:154:4: W0621: Redefining name 'product_to_build' from outer scope (line 153) (redefined-outer-name)
build_product_images.py:170:4: R1705: Unnecessary "else" after "return" (no-else-return)
build_product_images.py:28:0: C0411: standard import "import argparse" should be placed before "import conf" (wrong-import-order)
build_product_images.py:29:0: C0411: standard import "import subprocess" should be placed before "import conf" (wrong-import-order)
```

build_product_images.py
```diff
@@ -81,32 +101,41 @@ def build_image_tags(image_name, image_version, product_version):
...
+        "--platform",
+        ",".join(args.architecture),
+    ]
+
     if args.push:
-        commands.append(["docker", "push", "--all-tags", image_name])
+        commands.append(
+            "--push",
+        )
+    if not args.push and len(args.architecture) == 1:
```
sbernauer, line 131: Why not loud the images into local registry when build e.g. 2 images? (producing the following warning as well)
```
WARNING: No output specified for docker-container driver. Build result will only remain in the build cache. To push result image into registry use --push or to load image into docker use --load
```

sbernauer (changes requested):
Thanks for your work! First of your provided Image works for me, e.g. `docker run docker.stackable.tech/maleware/zookeeper:3.8.0-stackable0` chooses correct platform and runs.

Some minor thing in the Python script commented.

Sorry for my stupid question, but is my amd64 machine supposed to be able to build arm64 images? I guess not because without true virtualisation my architecture remains inside the build container and buildx is not using something like quemu(?).

Anyhow for the record when at least trying it, I get the following error, i think this is expected. Building just amd64 works fine and produces a runnable image.
```
python3 build_product_images.py -p zookeeper -v 3.8.0 -i 0.42.0 -a linux/amd64 linux/arm64
builder
WARNING: No output specified for docker-container driver. Build result will only remain in the build cache. To push result image into registry use --push or to load image into docker use --load
[+] Building 26.1s (10/36)                                                                                                                                                              
 => [internal] booting buildkit                                                                                                                                                    2.0s
 => => pulling image moby/buildkit:buildx-stable-1                                                                                                                                 1.4s
 => => creating container buildx_buildkit_builder0                                                                                                                                 0.5s
 => [internal] load build definition from Dockerfile                                                                                                                               0.0s
 => => transferring dockerfile: 2.56kB                                                                                                                                             0.0s
 => [internal] load .dockerignore                                                                                                                                                  0.0s
 => => transferring context: 2B                                                                                                                                                    0.0s
 => [linux/arm64 internal] load metadata for docker.stackable.tech/stackable/java-base:11-stackable0                                                                               1.5s
 => [linux/amd64 internal] load metadata for docker.stackable.tech/stackable/java-base:11-stackable0                                                                               1.5s
 => [internal] load build context                                                                                                                                                  0.1s
 => => transferring context: 21.57MB                                                                                                                                               0.1s
 => [linux/arm64  1/15] FROM docker.stackable.tech/stackable/java-base:11-stackable0@sha256:fd428a260606ab9047ca67cf3bceba5ac3859c35fd63feeda014e7523b2ccacc                      22.2s
 => => resolve docker.stackable.tech/stackable/java-base:11-stackable0@sha256:fd428a260606ab9047ca67cf3bceba5ac3859c35fd63feeda014e7523b2ccacc                                     0.0s
 => => sha256:8d0b9537f40b879c253845da327ef7e0a256a61c5325f8e86b2ed9e8c99babdb 3.73kB / 3.73kB                                                                                     0.0s
 => => sha256:7d1251b33ec1adfa3286195ae714bccec76675fabc753a7a8128215a48f218dc 17.14kB / 17.14kB                                                                                   0.0s
 => => sha256:4f8ddd7f5a755f537dd9d5f553c8c78171dcf3018c5fc96676a07380d3e14e20 1.75kB / 1.75kB                                                                                     0.1s
 => => sha256:ef1664cb8bf1485d3e10bf8cfc0086da606f6387f9955e22279b3bcde388e46e 110.71MB / 110.71MB                                                                                20.7s
 => => sha256:54e56e6f85721741ee7bf0336de8ad3bf138a56769a6d0097b600a0e361be58d 39.62MB / 39.62MB                                                                                  10.2s
 => => extracting sha256:54e56e6f85721741ee7bf0336de8ad3bf138a56769a6d0097b600a0e361be58d                                                                                          0.5s
 => => extracting sha256:4f8ddd7f5a755f537dd9d5f553c8c78171dcf3018c5fc96676a07380d3e14e20                                                                                          0.0s
 => => extracting sha256:ef1664cb8bf1485d3e10bf8cfc0086da606f6387f9955e22279b3bcde388e46e                                                                                          1.5s
 => => extracting sha256:8d0b9537f40b879c253845da327ef7e0a256a61c5325f8e86b2ed9e8c99babdb                                                                                          0.0s
 => => extracting sha256:7d1251b33ec1adfa3286195ae714bccec76675fabc753a7a8128215a48f218dc                                                                                          0.0s
 => [linux/amd64  1/15] FROM docker.stackable.tech/stackable/java-base:11-stackable0@sha256:fd428a260606ab9047ca67cf3bceba5ac3859c35fd63feeda014e7523b2ccacc                       0.0s
 => => resolve docker.stackable.tech/stackable/java-base:11-stackable0@sha256:fd428a260606ab9047ca67cf3bceba5ac3859c35fd63feeda014e7523b2ccacc                                     0.0s
 => ERROR [linux/arm64  2/15] RUN microdnf update &&     microdnf install tar gzip zip &&     microdnf install shadow-utils openssl &&     microdnf clean all                      0.3s
 => CANCELED [linux/amd64  2/15] RUN microdnf update &&     microdnf install tar gzip zip &&     microdnf install shadow-utils openssl &&     microdnf clean all                   0.3s
------                                                                                                                                                                                  
 > [linux/arm64  2/15] RUN microdnf update &&     microdnf install tar gzip zip &&     microdnf install shadow-utils openssl &&     microdnf clean all:                                 
#0 0.071 .buildkit_qemu_emulator: /bin/bash: Invalid ELF image for this architecture                                                                                                    
------
Dockerfile:17
--------------------
  16 |     
  17 | >>> RUN microdnf update && \
  18 | >>>     microdnf install tar gzip zip && \
  19 | >>>     microdnf install shadow-utils openssl && \
  20 | >>>     microdnf clean all
  21 |     
--------------------
error: failed to solve: process "/dev/.buildkit_qemu_emulator /bin/bash -o pipefail -c microdnf update &&     microdnf install tar gzip zip &&     microdnf install shadow-utils openssl &&     microdnf clean all" did not complete successfully: exit code: 255
Traceback (most recent call last):
  File "build_product_images.py", line 230, in <module>
    main()
  File "build_product_images.py", line 222, in main
    run_commands(args.dry, commands)
  File "build_product_images.py", line 150, in run_commands
    subprocess.run(cmd, check=True)
  File "/usr/lib/python3.8/subprocess.py", line 516, in run
    raise CalledProcessError(retcode, process.args,
subprocess.CalledProcessError: Command '['docker', 'buildx', 'build', '--build-arg', 'PRODUCT=3.8.0', '-t', 'docker.stackable.tech/stackable/zookeeper:3.8.0-stackable0.42.0', '-f', 'zookeeper/Dockerfile', '--platform', 'linux/amd64,linux/arm64', '.']' returned non-zero exit status 1.
```

sbernauer:
Disregard my last comment, I should have read your description first.
Trying with only a single `-a` as follows works fine :+1: 
```
$ python3 build_product_images.py -p zookeeper -v 3.8.0 -i 0.42.0 -a linux/arm64
```

build_product_images.py
Maleware: Would be fine by me, works on my machine too.

build_product_images.py
Maleware: Sadly manifest lists are currently not supported by buildx to be loaded into local registry.  You can add -d and put a --load behind a multi arch build (remove --push, both together are not working too), buildx will complain that --load can't be used for multi arch images.

sbernauer: Alright thx!

build_product_images.py
Maleware: Resolved some of them, thanks for your hint.

sbernauer (dismissed):
thx!

build_product_images.py
```diff
@@ -60,6 +60,11 @@ def parse_args():
         required=True,
         type=check_architecture_input
     )
+    parser.add_argument(
+        "--path",
```
razvan, line 64: According to the docker docs, this is the _organization namespace_ [1] so I would call it `--organization` and also add the short option `-o`

[1] https://docs.docker.com/glossary/#organization-name

Maleware: Done

razvan (approved):
lgtm

PR #159, opened 2022-08-09
Remove openssl from zookeeper image

maltesander:
https://github.com/stackabletech/zookeeper-operator/pull/533 uses the tools image for the initcontainer, therefore openssl is no longer needed here.

### DIFF
--- a/zookeeper/Dockerfile
+++ b/zookeeper/Dockerfile
@@ -16,7 +16,7 @@ SHELL ["/bin/bash", "-o", "pipefail", "-c"]
 
 RUN microdnf update && \
     microdnf install tar gzip zip && \
-    microdnf install shadow-utils openssl && \
+    microdnf install shadow-utils && \
     microdnf clean all
 
 COPY zookeeper/stackable /stackable


teozkr:
This will break old versions of the operator, so I'd put this on hold until we have a release with https://github.com/stackabletech/zookeeper-operator/pull/533.

sbernauer:
Wouldn't the correct procedure be to add a Changelog entry with a breaking change and release a new minor version [for each zk version]?
The new ZK-PR can than use the new minor version.

maltesander:
> This will break old versions of the operator, so I'd put this on hold until we have a release with [stackabletech/zookeeper-operator#533](https://github.com/stackabletech/zookeeper-operator/pull/533).

Yeah thats fine with me. It will break older operators once we have a new tag (either zookeeper version or fix here).
Priority is not very high, but i think we should use tools image as far as possible for stuff like this.

lfrancke:
Not sure why this is on hold

sbernauer:
We didn't want to release a breaking change just for this openssl removal. The idea was to wait for some other breaking change and release them together

soenkeliebau:
I'll check on whether this is still relevant in ligth of the decision to remove the tools images.

soenkeliebau:
If we go ahead with the decision to remove tools images and instead use product images for init containers as well we can close this ticket.
I was unable to find the issue where this was discussed to be honest, but I remember most of us being in favor of removing the tools image because this created an unnecessary dependency between operators.

lfrancke:
Okay, we made the decision to remove tools image correct? In that case this can be closed?

maltesander:
Yeah not relevant anymore then.

PR #162, opened 2022-08-11
Custom Python image for integration tests

razvan:
Part of : https://github.com/stackabletech/issues/issues/230


testing-tools/CHANGELOG.md
```diff
@@ -0,0 +1,9 @@
+# Changelog
+
+## [Unreleased]
+
+### Added
+
+- Initial version with Java, Python and all packages currently installed by kuttl tests at runtime ([#xxx]).
+
+[#xxx]: https://github.com/stackabletech/docker-images/pull/xxx
```
sbernauer, line 9:
```suggestion
- Initial version with Java, Python and all packages currently installed by kuttl tests at runtime ([#162]).

[#162]: https://github.com/stackabletech/docker-images/pull/162
```

sbernauer (commented):
Could you please also add `trino==0.315.0 ` for https://github.com/stackabletech/trino-operator/blob/main/tests/templates/kuttl/smoke/21-install-requirements.yaml?

sbernauer:
Could we also add `kafka-python3==3.0.0`?

sbernauer:
And also the libs from https://github.com/stackabletech/superset-operator/blob/main/tests/templates/kuttl/smoke/requirements.txt?

sbernauer:
Having `pandas==1.4.3` would be great as we use that for the demos and we could re-use the image

sbernauer:
Could you please also add the libs from https://github.com/stackabletech/superset-operator/blob/main/tests/templates/kuttl/smoke/requirements.txt like idna and requests?

sbernauer (approved):
It would be great if you would find the time to test the new image for all the product-tests we want to switch to this image.
If you don't have the time that's ok as well.

PR #169, opened 2022-09-13
spark-k8s and pyspark-k8s: Add s3a and abfs libs

### LINKED ISSUE #172: spark-k8s and pyspark-k8s: Add s3a and abfs libs
S3 libs are needed for basically everything (e.g. reading/writing data and history-server) therefore we want to pre-ship them with our images.
Another reason is to put *the correct* versions in the image, as many users will be confused on which versions to put there

Implementation can be copied from hive image

- [x] Must: Libs added to new Image version. Only adding it to Spark 3.3.0 only is ok, as older Spark versions have older Hadoop versions and need other aws libs
- [ ] Must: kuttl test added
- [ ] Could: kuttl tests using s3a libs via pvcs or similar are removed
- [x] Documentation & changelog updated to mention the inclusion of these libs

sbernauer:
Fixes #172

### DIFF
--- a/conf.py
+++ b/conf.py
@@ -127,13 +127,17 @@ products = [
         'versions': [
             {
                 'product': '3.2.1',
-                'hadoop': '3.2',
                 'python': '39',
+                'hadoop_short_version': '3.2',
             },
             {
                 'product': '3.3.0',
-                'hadoop': '3',
                 'python': '39',
+                'hadoop_short_version': '3',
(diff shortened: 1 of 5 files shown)


razvan (approved):

sbernauer:
Adding tests for this in the operator is tracked in https://github.com/stackabletech/docker-images/issues/172

Issue #170, opened 2022-09-13
ubi8-minimal microdnf --enablerepo=ubi-8-appstream can't be used currently


Maleware:
The current situation for e.g. ubi8-rust-builder:
we use `FROM registry.access.redhat.com/ubi8/ubi-minimal AS builder` as base image, to install stuff we use `&& microdnf install --disablerepo=* --enablerepo=ubi-8-appstream-rpms --enablerepo=ubi-8-baseos curl findutils gcc gcc-c++ make cmake openssl-devel pkg-config systemd-devel unzip pearl -y \
  && rm -rf /var/cache/yum`
  
  This leads to the error that ubi-8-appstream is not found

Maleware:
The problem was solved by adding `--enablerepo=ubi-8-appstream-rpms --enablerepo=ubi-8-baseos-rpms` instead of `--enablerepo=ubi-8-appstream --enablerepo=ubi-8-baseos`

vsupalov:
The problem also does not show up if the base image is pinned to `registry.access.redhat.com/ubi8/ubi-minimal:8.6-902`

lfrancke:
Thank you.
I'd like to understand
- why this happened
- what we need to do to fix it _now_
- what we can do to prevent it in the future if there is anything
- if this was announced somewhere (related to the previous item) so we can follow that in the future

Maleware:
To the best of my knowledge:

- This happened because RH decided without further notice to switch their repo name from `ubi-8-appstream` to `ubi-8-appstream-rpms` . This was breaking for ubi8-rust-builder. Building product images would have worked as long as we do not touch the rust builder as it currently is
- To fix it, we can rename to `icrodnf update --disablerepo=* --enablerepo=ubi-8-appstream-rpms --enablerepo=ubi-8-baseos-rpms -y` but we do not know if and when RH decides to go back to the old thing. 
- to the best of my knowledge I haven't seen anything announcing this change, but investigating now.

Maleware:
Fix through: https://github.com/stackabletech/docker-images/pull/171

lfrancke:
Most of this comes from me from about a year ago and I copied it straight from the docs. Can one of you see what the docs say now?

Maleware:
Don't think so, the documentation still states you should use ubi-8-appstream. Looks exactly like what we are doing in both links. We just use microdnf instead of yum and that's what they want us to do.

Maleware:
[...]

PR #171, opened 2022-09-13
Enable Renovate to manage pinning of base image versions

Maleware:
Fixes https://github.com/stackabletech/docker-images/issues/170
Updating repo name due to change of RH

We are not tagging the docker image versions in this PR, as we want Renovate to manage this for us. So after this PR we will briefly have a state where nothing is pinned/tagged, until Renovate creates a follow up PR which pins all images.

### DIFF
--- a/java-base/Dockerfile
+++ b/java-base/Dockerfile
@@ -1,4 +1,4 @@
-FROM registry.access.redhat.com/ubi8/ubi-minimal:8.5
+FROM registry.access.redhat.com/ubi8/ubi-minimal:8.6-941
 
 ARG PRODUCT
 ARG RELEASE="1"
--- a/java-base/Dockerfile
+++ b/java-base/Dockerfile
@@ -1,4 +1,4 @@
-FROM registry.access.redhat.com/ubi8/ubi-minimal:8.6-941
+FROM registry.access.redhat.com/ubi8/ubi-minimal
 
 ARG PRODUCT
 ARG RELEASE="1"
--- a/renovate.json
+++ b/renovate.json
@@ -1,0 +1,32 @@
+{
+  "extends": [
+    "helpers:pinGitHubActionDigests",
+    "docker:pinDigests"
+  ],
+  "labels": [
+    "dependencies"
+  ],
+  "prCreation": "not-pending",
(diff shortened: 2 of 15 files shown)


renovate.json
```diff
@@ -0,0 +1,32 @@
+{
```
soenkeliebau, line 1: paging @lfrancke or @teozkr as the resident Renovate experts to have a look :)

soenkeliebau: But would't it replace the tag with the digest and then add the tag in a comment at the end like it does with github actions?

https://github.com/stackabletech/operator-templating/pull/161/files

teozkr: Let me go check...

teozkr: https://docs.renovatebot.com/docker/#digest-updating implies that it will try to update both, but I'm not sure about to which degree this works to for third-party registries (especially "weird" ones that don't use Docker's own registry server).

I guess we could try running it locally and see what it tries to do...

soenkeliebau: Further [down](https://docs.renovatebot.com/docker/#digest-updating)) it sounds like it would not update though..

> To keep things simple, Renovate keeps the Docker tag in the FROM line, like this: FROM node:14.15.1@sha256:d938c1761e3afbae9242848ffbb95b9cc1cb0a24d889f8bd955204d347a7266e.
> 
> Digest Updating
> If you follow our advice to replace a simple tag like node:14 with a pinned digest node:14@sha256:d938c1761e3afbae9242848ffbb95b9cc1cb0a24d889f8bd955204d347a7266e, you will get Renovate PRs whenever the node:14 image is updated on Docker Hub.

java-base/Dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM registry.access.redhat.com/ubi8/ubi-minimal:8.5
+FROM registry.access.redhat.com/ubi8/ubi-minimal:8.6-941
```
teozkr, line 1: Seems like this introduced a new tag?

soenkeliebau: duh :(
This was missing the trailing " AS" that my search and replace expression contained.

PR #179, opened 2022-09-22
Patch repo ubi 8 appstream and base os

Maleware:
Pin all versions with digests, checkout why renovate is not pinning digests on it's own. Until then we should have a solution and pin the current version to protect us from sudden changes.


soenkeliebau:
I have pushed a commit to update all ubi8-minimal images to tag 8.6.
This should cause renovate to still update them, but keep the tag in the docker string, which will make it easier to see what we refer to going forward (as opposed to just a digest being there, which is kinda hard to resolve back to a version).

soenkeliebau (approved):

PR #230, opened 2022-10-21
hbase: include hadoop native libs

adwk67:
Add hadoop native libs to image to so that compression (lz4, bzip2) can be applied to tables.
Don't merge - this is for compression tests only at the moment.

##### TODOs
- hadoop version is currently hard-coded as the latestone (3.3.4). Do we want to add this to the argument matrix, or are the native libs static enough for it not to matter?
- currently this adds the whole hadoop install, which makes the image really large. Only the native libs should be added.

### DIFF
--- a/conf.py
+++ b/conf.py
@@ -46,22 +46,27 @@ products = [
             {
                 'product': '2.4.6',
                 'phoenix': '2.4-5.1.2',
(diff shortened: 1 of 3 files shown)


adwk67:
Superseded by changes in branch `compression_test_malte`

PR #252, opened 2022-11-10
Removed versions that are outdated 

maltesander:
Removed obsolete supported OPA and NiFi versions which are not listed under supported-versions.adoc in the operator repos anymore.


sbernauer (changes requested):
What place are you referring to?
I could only find https://github.com/stackabletech/opa-operator/blob/main/docs/modules/ROOT/partials/supported-versions.adoc, which states we do support 0.27.1 - 0.45.0

Additionally question is how long do we support platform releases? I don't think we have a strategy on that yet. I don't want to be the guy deleting images that are used in production somewhere. I know this PR does not delete Images, but if we support the Image i would prefer to keep them in this list.
Anyhow, i would suggest speaking about this in the Arch-Meeting

maltesander:
> What place are you referring to? I could only find https://github.com/stackabletech/opa-operator/blob/main/docs/modules/ROOT/partials/supported-versions.adoc, which states we do support 0.27.1 - 0.45.0

Yeah i dont know where i saw that these opa versions are not listed... The NiFi one at least should be right. Reverted the OPA changes. 

And yeah i think we should talk about how long we support the images / versions.

sbernauer:
If it's ok i would approve when discussion did take place to not accidentally merge this?

sbernauer (approved):
As NiFi 1.13.2 is not documented as supported anywhere LGTM

PR #264, opened 2022-11-30
Updated with Trino dep + slimmed image size

### LINKED ISSUE #168: Airflow: Reduce image size by 200MB
Pulling the airflow image i noticed that the last layer (200MB) is duplicated. This results in a unnecessary big image and longer download times.
image
The command that duplicated the whole layer seems to be `chown -R stackable:stackable /stackable` [here](https://github.com/stackabletech/docker-images/blob/968347cad178be22afd896dc0ea943a8824a02c3/airflow/Dockerfile#L83)

DoD:
- [ ] Remove 200MB from the airflow image

### LINKED ISSUE #263: Request - Installing Trino Provider Dependencies
## Description
I was exploring an use-case where we want to use Trino in our DAGs. Looking into how Airflow runs and how the dependencies are handled, I have reach the possible conclusion it might quite easy to add with the way you have created the docker images.

I have build and run it locally with the added dependency and was able to configure the Trino connection and run simple queries against a test Trino cluster, which by the looks of it, works.

zaultooz:
Pull request for adding the Trino provider. I changed the ordering on docker cmds to reduce the size of the final image. Locally it was about 1 GB smaller.

Closes #168 
Closes #263

### DIFF
--- a/airflow/Dockerfile
+++ b/airflow/Dockerfile
@@ -7,7 +7,7 @@ COPY airflow/constraints-${PRODUCT}-python${PYTHON}.txt /tmp/constraints.txt
 
 # the mysql provider is currently excluded.
 # Requires implementation of https://github.com/apache/airflow/blob/2.2.5/scripts/docker/install_mysql.sh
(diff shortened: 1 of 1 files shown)


sbernauer (approved):
Awesome, thanks for your contribution!
`python3 ./build_product_images.py -p airflow -i 0.42.0 -v 2.4.1 -a linux/amd64` did pass and the image worked

sbernauer:
Version `2.4.1-stackable0.6.0` including this change should be available now

zaultooz:
Cool to hear, thanks! I give the new version a spin.

PR #270, opened 2022-12-09
Moved statsd-exporter to main image

maltesander:
Now copying statsd-exporter from  its docker image to the main image for airflow and superset (hopefully multi arch compatible)

@Maleware can you confirm :) ?

### DIFF
--- a/superset/Dockerfile
+++ b/superset/Dockerfile
@@ -57,7 +57,7 @@ RUN cat /tmp/patches/* | patch \
     --strip=1
 
 # Final image
-
+FROM prom/statsd-exporter:0.3.0@sha256:a9c27602d6f6b86527657922b6a87c12789f7f9b39a90f1513e8c665c941f26a as statsd-exporter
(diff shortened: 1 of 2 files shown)


superset/Dockerfile
```diff
@@ -57,7 +57,7 @@ RUN cat /tmp/patches/* | patch \
     --strip=1
 
 # Final image
-
+FROM prom/statsd-exporter:0.3.0@sha256:a9c27602d6f6b86527657922b6a87c12789f7f9b39a90f1513e8c665c941f26a as statsd-exporter
```
sbernauer, line 60: I would add that before the comment `# Final image`

superset/Dockerfile
```diff
@@ -95,6 +95,7 @@ RUN microdnf update \
 
 COPY superset/licenses /licenses
 COPY --from=builder /stackable ${HOME}
+COPY --from=statsd-exporter /bin/statsd_exporter /stackable/statsd_exporter
```
sbernauer, line 98: Nit: When you already touch the Superset Image it would be great to include the fix from https://github.com/stackabletech/docker-images/pull/264/files to get rid of the `chown -R stackable:stackable /stackable`

sbernauer (approved):

sbernauer (approved):
Thx!

PR #291, opened 2023-01-20
Add Vector to the Airflow images

siegfriedweber:
Add Vector to the Airflow images

Prerequisite for stackabletech/airflow-operator#213

### DIFF
--- a/airflow/Dockerfile
+++ b/airflow/Dockerfile
@@ -1,4 +1,4 @@
-FROM docker.stackable.tech/stackable/vector:0.26.0-stackable1.0.0@sha256:4be0d98ca353b3ddb54212ced702504d58995b2b75fbfa07fe001fd992513692 AS airflow-build-image
+FROM docker.stackable.tech/sandbox/logging/vector:0.26.0-stackable23.4.0-rc1@sha256:d350ad8da54acec6e285130911762135cbe16cbc307d2ae977e6dd2fc1980d23 AS airflow-build-image
 
 ARG PRODUCT
 ARG PYTHON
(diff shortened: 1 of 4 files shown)


maltesander (approved):
LGTM.

airflow/Dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM registry.access.redhat.com/ubi8/ubi-minimal:8.6@sha256:c5ffdf5938d73283cec018f2adf59f0ed9f8c376d93e415a27b16c3c6aad6f45 AS airflow-build-image
```
nightkr, line 1: Should we move vector out of the sandbox now that we have "top-level" images depending on it?

siegfriedweber: Yes, this was on my TODO list before merging. Usually I add this TODO list to the description of the pull request but I forgot it this time.

airflow/Dockerfile
```diff
@@ -60,32 +60,24 @@ RUN microdnf install -y yum python${PYTHON} \
 ENV HOME=/stackable
 ENV AIRFLOW_USER_HOME_DIR=/stackable
 ENV PATH=$PATH:/bin:$HOME/app/bin
-ENV AIRFLOW_UID="50000"
+ENV AIRFLOW_UID="1000"
```
nightkr, line 63: Does this require a manual migration step for users?

maltesander (approved):
LGTM.

airflow/Dockerfile
siegfriedweber: The data should not be affected by this change. But the UID stored in the airflow-operator must be in sync with this one. So the airflow-operator 23.4.0 is only compatible with the stackable23.4.0 Airflow image.

nightkr: Alright, fair enough.

nightkr (approved):

PR #302, opened 2023-01-30
Allow -rc[0-9]+ tags to images.

razvan:
Generate images with release candidate tags such as `23.4.0-rc1`

In addition, images are only tagged with one tag instead of two.

Until now,  tagging with  "23.1.0" also produced images with the tag "23.1" . This second so called  "platform" tag has been removed.

Part of: https://github.com/stackabletech/issues/issues/324


maltesander (approved):
LGTM.

PR #304, opened 2023-01-30
fix tags

razvan:
- Revert "Replace -rc1 with .rc1 to prevent GH actions from crashing. (#303)"
- Move tag validation to the build script.


maltesander (approved):
LGTM!

PR #321, opened 2023-02-06
Build images using docker buildx bake

nightkr:
This PR changes the build workflow to generate a Bakefile (https://docs.docker.com/build/bake/), which is then passed to a single `docker buildx` call. This allows us to refer to other Dockerfiles in the tree directly, rather than having to build every step in the dependency chain manually, updating all the image references in between.

Additionally, this lets Docker build and push multiple the whole tree in parallel, rather than building the images one by one.

~~Currently Airflow and Zookeeper (and their dependencies) have been updated to take advantage of local dependencies. Other images may not work without updating to the `docker/dockerfile:1` syntax.~~ All images have now been updated.

### DIFF
--- a/build_product_images.py
+++ b/build_product_images.py
@@ -139,9 +139,9 @@ def generate_bakefile(args: Namespace):
     """
     targets = {}
     groups = {}
-    products = {product["name"]: product for product in conf.products}
-    product_names = list(products.keys())
-    for product_name, product in products.items():
+    product_names = [product["name"] for product in conf.products]
+    for product in conf.products:
+        product_name = product["name"]
         product_targets = {}
         for version_dict in product.get("versions"):
             product_targets.update(bakefile_product_version_targets(args, product_name, version_dict, product_names))
--- a/build_product_images.py
+++ b/build_product_images.py
@@ -33,6 +33,7 @@ from argparse import Namespace, ArgumentParser
 import subprocess
 import conf
 import re
+import json
 
 # This is the stackable release version
 DEFAULT_IMAGE_VERSION_FROMATS = [re.compile("[2-9][0-9]\.[1-9][0-2]?\.\d+"), re.compile("[2-9][0-9]\.[1-9][0-2]?\.\d+-rc[1-9]\d?")]
@@ -101,19 +102,17 @@ def build_image_args(version, release_version):
     - version: Can be a str, in which case it's considered the PRODUCT
                 or a dict.
     """
-    result = []
+    result = {}
 
     if isinstance(version, dict):
         for k, v in version.items():
-            result.extend(["--build-arg", f"{k.upper()}={v}"])
-        result.extend(["--build-arg", f"RELEASE={release_version}"])
+            result[k.upper()] = v
+        result["RELEASE"] = release_version
     elif isinstance(version, str) and isinstance(release_version, str):
-        result = [
-            "--build-arg",
-            f"PRODUCT={version}",
-            "--build-arg",
-            f"RELEASE={release_version}",
-        ]
+        {
+            "PRODUCT": version,
+            "RELEASE": release_version,
+        }
     else:
         raise ValueError(f"Unsupported version object: {version}")
 
@@ -125,10 +124,8 @@ def build_image_tags(image_name: str, image_version: str, product_version: str) 
     Returns the --tag command line arguments that are used by the docker build command.
     """
     return [
-        "-t",
         f"{image_name}:{product_version}-stackable{image_version}",
     ]
-
 
 def build_and_publish_image(args: Namespace, product_name: str, versions: Dict[str, str]) -> List[List[str]]:
     """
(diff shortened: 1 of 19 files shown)


conf.py
```diff
@@ -28,56 +33,70 @@
...
-            },
+            # Hadoop 2.10.1 is no longer supported
+            # {
+            #     'product': '2.3.9',
+            #     'hadoop': '2.10.1',
+            #     'jackson_dataformat_xml': '2.7.9',
+            #     'aws_java_sdk_bundle': '1.11.271',
+            #     'azure_storage': '7.0.1',
+            #     'azure_keyvault_core': '1.0.0',
+            # },
```
nightkr, line 96: I assume that it's an oversight that we still have this in the build file, are we OK dropping it?

conf.py
```diff
@@ -204,47 +246,48 @@
...
-    {
-        'name': 'testing-tools',
-        'versions': [{'product': '0.1.0'}],
+        'versions': [{'product': '0.2.0', 'stackable-base': '1.0.0'}],
     },
+    # Build is broken
+    # {
+    #     'name': 'testing-tools',
+    #     'versions': [{'product': '0.1.0'}],
+    # },
```
nightkr, line 282: This seems to have bitrotted heavily, do we still want this?

conf.py
```diff
@@ -204,47 +246,48 @@
...
     },
     {
         'name': 'tools',
-        'versions': [{'product': '0.2.0'}],
-    },
-    {
-        'name': 'testing-tools',
-        'versions': [{'product': '0.1.0'}],
+        'versions': [{'product': '0.2.0', 'stackable-base': '1.0.0'}],
     },
```
nightkr, line 277: Does anything still depend on tools are we OK dropping it?

nightkr:
The Hadolint DL3006 warnings are false positives, the `stackable/image/` paths are replaced with other in-tree images by Bake, so it is correct that they don't have a version tag.

conf.py
Maleware: As far as I know, we dropped all dependencies to tools

conf.py
razvan: Oh, I didn't know that. This image is used pretty much everywhere in the kuttl tests.

sbernauer: Yeah, the tools Image should not be needed anymore, but the testing-tools is used in tests - obviously - and demos

build_product_images.py
```diff
@@ -125,48 +127,97 @@ def build_image_tags(image_name: str, image_version: str, product_version: str)
...
-    Returns a list of commands that need to be run in order to build and
-    publish product images.
+    Generates a Bakefile (https://docs.docker.com/build/bake/file-definition/) describing how to build the whole image graph.
 
-    For local building, builder instances are supported.
+    build_and_publish_images() ensures that only the desired images are actually built.
+    """
+    targets = {}
+    groups = {}
+    contexts = {}
```
razvan, line 142: "contexts" is not used

build_product_images.py
```diff
@@ -125,48 +127,97 @@ def build_image_tags(image_name: str, image_version: str, product_version: str)
...
-        commands.append(
-            "--push",
-        )
-    if not args.push and len(args.architecture) == 1:
-        commands.append(
-            "--load",
-        )
 
-    commands.append(".")
+def build_and_publish_image(args: Namespace, product_name: str, bakefile) -> List[List[str]]:
```
razvan, line 193:
```suggestion
def build_and_publish_image(args: Namespace, product_name: str, bakefile) -> List[Dict[str, Sequence[str]]]:
```

build_product_images.py
```diff
@@ -125,48 +127,97 @@ def build_image_tags(image_name: str, image_version: str, product_version: str)
...
 
-    For local building, builder instances are supported.
+    build_and_publish_images() ensures that only the desired images are actually built.
+    """
+    targets = {}
+    groups = {}
+    contexts = {}
+    products = {product["name"]: product for product in conf.products}
+    product_names = list(products.keys())
+    for product_name, product in products.items():
```
razvan, line 145:
```suggestion
    product_names = [p.get("name") for p in conf.products]
    for product_name, product in zip(product_names, conf.products):
```

razvan (dismissed):
Nice and elegant. Added some suggestions.

build_product_images.py
nightkr: Yeah, this was an old attempt to make dependency inference even more automatic. Away it goes!

build_product_images.py
nightkr: I think I'd rather just remove the return type entirely then, the dict doesn't really have a meaningful "shape" that we can reasonably type like this (at least not without defining it as a class/namedtuple instead).

build_product_images.py
nightkr: This is an old leftover from the context mechanism, I'll go simplify.

build_product_images.py
nightkr: 1c5776f

build_product_images.py
nightkr: 64f35d6

razvan (dismissed):
nice

conf.py
nightkr: Ok, I've created https://github.com/stackabletech/docker-images/issues/323 since that's a breaking change that I don't really want to intermingle with this PR.

conf.py
nightkr: Ok, I've created https://github.com/stackabletech/docker-images/issues/324 for this then

conf.py
nightkr: Reenabled in cc8e73a605db1450a6b60f7f3d7d2deed63aaeb1.

razvan (approved):

PR #326, opened 2023-02-08
Unpin testing-tools dependencies

### LINKED ISSUE #324: Fix testing-tools dockerfile bitrot

nightkr:
Debian seems to prune old versions of packages from their repository regularly. After discussing with @razvan I decided to remove the pins, since updating the bounds would just cause the problem to return once the new versions were also pruned, meaning that we still wouldn't have reliable reproducibility.

Fixes #324.


sbernauer (approved):
LGTM for me

PR #327, opened 2023-02-09
pyspark-k8s: decouple product from java version

adwk67:
Currently we use the product tag (e.g. `3.3.0`) to

- name the image e.g. `3.3.0-stackable23.4.0-rc1`
- identify which archive to use e.g. `spark-3.3.0-bin-hadoop3.tgz`

which means it is not possible to specify and tag the same spark version against different versions of scala (with which spark is compiled) or java (on which it should run). This PR changes the tag usage so that:
- `PRODUCT` is used just for labelling (and can therefore be qualified with java and/or scala versions e.g. `3.3.0-java17-...`)
- `JAVA` refers to the installed JVM

This can extended when needed to add `SCALA` as an argument.


razvan (approved):
Fine, but just for the record, I really feel this should have been a different "product".

PR #328, opened 2023-02-09
Add Vector to the Superset images

siegfriedweber:
Add Vector to the Superset images

Prerequisite for stackabletech/superset-operator#317


maltesander (approved):
LGTM!

PR #334, opened 2023-02-23
Feat/preflight check

fhennig:


build_product_images.py
```diff
@@ -264,6 +265,14 @@ def check_architecture_input(architecture):
     return architecture
 
 
+def get_tags_for_product(bakefile, product: str) -> List[str]:
+    tags = []
+    targets = bakefile["group"][product]["targets"]
+    for t in targets:
+        tags.extend(bakefile["target"][t]["tags"])
```
nightkr, line 272: `product` can actually also be a target (for example: "I only want to build Hadoop 3.3.4") or a meta-group ("I want to build all versions of all products"). I think a more correct answer would be something like:
```suggestion
    for target in bakefile["group"].get(product, {}).get("targets", []):
        tags.extend(get_tags_for_product(bakefile, target))
    tags.extend(bakefile["target"].get(product, {}).get("tags", [])
```

nightkr: It might also be more correct to name the function something like `get_tags_for_target`.

fhennig: I see!

razvan:
Alternative PR: https://github.com/stackabletech/docker-images/pull/339

fhennig:
superseded by the alternative PR linked above

PR #338, opened 2023-02-28
Add krb5 image

nightkr:


krb5/Dockerfile
```diff
@@ -0,0 +1,10 @@
+# syntax=docker/dockerfile:1
+# RedHat UBI does not package krb5-server, so we cannot use it as a KDC
+# FROM stackable/image/stackable-base
+FROM rockylinux:8.7@sha256:68bef3459bbb8c33841575a7f71c4de94718b7bbd103fd0417a537395d40c592
```
sbernauer, line 4: Paging our OpenShift experts @razvan and @adwk67. Is there any chance we can get this OpenShift certified?

nightkr: No, but this will only be used for tests and maybe demos. We'll need to figure out if/how we want to ship a KDC at some point though…

(and how to document this in the meantime)

sbernauer: I was thinking we need this as a core service such as ticket server or similar from the beginning.
Works for me.

sbernauer (approved):

krb5/Dockerfile
nightkr: Well customers need *a* KDC (if they use Kerberos), but it doesn't have to be using our image.

nightkr: Added a note in 1f3dfe240c31adb029eec62ab601532a254fc005

sbernauer (approved):
lgtm

PR #352, opened 2023-03-28
hbase: Add krb5-workstation and openssl packages

sbernauer:


sbernauer:
Not needed any more, we already ungraded to 2.6

PR #360, opened 2023-04-20
Run preflight checks when tagging and merging to main

razvan:
- Submit OpenShift preflight checks to certification projects.
- Added preflight GH workflow.


fhennig (approved):
LGTM

PR #366, opened 2023-04-26
Add kerberos image to release workflow

sbernauer:


razvan (commented):
If it's this something that we ship with the platform then we need to also create a certification project and add it to the preflight jobs.

sbernauer:
Maybe https://github.com/stackabletech/docker-images/blob/main/krb5/README.adoc clarifies this a bit. The plan - so far - is to only use it in tests (think of like minio, postgres or redis)
Updated dev.yaml as well, thx!

.github/workflows/dev.yaml
```diff
@@ -68,6 +69,7 @@ jobs:
           - hbase
           - hive
           - kafka
+          - krb5
```
razvan, line 72: This image doesn't use ubi8 so it won't pass the preflight checks. This can be removed,

.github/workflows/release.yml
```diff
@@ -73,6 +74,7 @@ jobs:
           - hbase
           - hive
           - kafka
+          - krb5
```
razvan, line 77: Also remove this.

.github/workflows/dev.yaml
sbernauer: You're absolutely right, thx!

.github/workflows/release.yml
sbernauer: You're absolutely right, thx!

razvan (approved):

sbernauer:
Thx for the review!

PR #370, opened 2023-05-16
Added hello world Dockerfile

fhennig:


hello-world/Dockerfile
```diff
@@ -0,0 +1,26 @@
...
+
+RUN microdnf update && \
+    microdnf clean all
+
+USER stackable
+WORKDIR /stackable
+
+RUN curl -L https://repo.stackable.tech/repository/packages/hello-world/hello-world-${PRODUCT}.jar -o hello.jar
+
+CMD ["java", "-jar", "hello.jar", "--spring.config.location=file:mount/config/"]
```
sbernauer, line 26: Nit
```suggestion
RUN curl -L https://repo.stackable.tech/repository/packages/hello-world/hello-world-${PRODUCT}.jar -o hello-world.jar

CMD ["java", "-jar", "hello-world.jar", "--spring.config.location=file:mount/config/"]
```

sbernauer, line 26: Any particular reason to use `file:mount/config/` and not `file:/stackable/config/`? I *think* this aligns better with our current operators

sbernauer (changes requested):

hello-world/Dockerfile
fhennig: Ah! Good thing you bring it up, because I got this from the Hive operator and there the configmap is mounted under /stackable/mount/config.

I was also slightly confused initially, and not quite sure why it is like that, I'm happy to switch it, but should I investigate further?

fhennig: https://github.com/stackabletech/hive-operator/blob/main/rust/crd/src/lib.rs#L32

sbernauer:
```
pub const STACKABLE_CONFIG_DIR: &str = "/stackable/config";
[...]
pub const STACKABLE_CONFIG_MOUNT_DIR: &str = "/stackable/mount/config";
```
=> We need to mount the ConfigMap (here at  "/stackable/mount/config"). As the ConfigMap is mounted readonly it the files can not be changed (and also should not). We have to run sed on the files, as we need to put in some secrets.
So we copy whatever is mounted to "/stackable/config", so that we can run sed on it and have the final config in "/stackable/mount/config". I don't think we need the complexity here, the important bit for me is that the **effective** config is below /stackable/config (though not all operators do so :see_no_evil: )

fhennig: Good point. I'll change it!

PR #380, opened 2023-05-26
trino: Bump opa-authorizer version for Trino 414

sbernauer:
So we get https://github.com/stackabletech/trino-opa-authorizer/issues/24 in


maltesander (approved):
LGTM!

PR #381, opened 2023-05-26
Support zookeeper 3.8.1

maltesander:


zookeeper/Dockerfile
```diff
@@ -29,9 +30,10 @@ WORKDIR /stackable
 RUN curl -L https://repo.stackable.tech/repository/packages/zookeeper/apache-zookeeper-${PRODUCT}-bin.tar.gz | tar -xzC . && \
     ln -s /stackable/apache-zookeeper-${PRODUCT}-bin /stackable/zookeeper
 
-RUN curl https://repo.stackable.tech/repository/packages/jmx-exporter/jmx_prometheus_javaagent-0.16.1.jar \
-    -o /stackable/jmx/jmx_prometheus_javaagent-0.16.1.jar && \
-    chmod -x /stackable/jmx/jmx_prometheus_javaagent-0.16.1.jar
+RUN curl https://repo.stackable.tech/repository/packages/jmx-exporter/jmx_prometheus_javaagent-${JMX_EXPORTER}.jar \
+    -o /stackable/jmx/jmx_prometheus_javaagent-${JMX_EXPORTER}.jar && \
+    chmod +x /stackable/jmx/jmx_prometheus_javaagent-${JMX_EXPORTER}.jar && \
```
sbernauer, line 35: Why the `+x`? I guess invoking the jar manually does not make sense, but interested in the reason :)

maltesander: Just for safety. It worked without for most products but i think failed for iirc Kafka.

sbernauer: Hmm I would *slightly* prefer not having executable but if it's needed for a product than we should go with consistency I guess

sbernauer (approved):

PR #396, opened 2023-06-21
Feature/multi arch workflow

Maleware:
This creates a Workflow with buildjet and builds multi-arch product images. Those get pushed in stackable-experimental.

docker buildx build --platform linux/amd64,linux/arm64 is actually not available. However, this command wouldn't make use of the possibility that two different native runners are available. 

This workflow is supposed to leverage the ability to build nativ on amd64 & arm64 and in the end build those two together into a manifest list and push it without the platform specific endings.


.github/workflows/dev_arm.yaml
```diff
@@ -0,0 +1,92 @@
+---
+name: Publish 0.0.0-dev images
```
razvan, line 2:
```suggestion
name: Publish ARM 0.0.0-dev images
```

image_tools/bake.py
```diff
@@ -114,6 +115,11 @@ def bakefile_product_version_targets(
     tags = build_image_tags(image_name, args.image_version, versions["product"])
     build_args = build_image_args(versions, args.image_version)
 
+    if args.multiarch:
+        if args.architecture[0] == "linux/amd64":
+            tags[0] = tags[0] + "-amd64"
+        else:
+            tags[0] = tags[0] + "-arm64"
```
razvan, line 122: I would move this to the `build_image_tags` function for cleanness.

razvan (commented):
Have you seen the actionlint errors that say that `buildjet-4vcpu-ubuntu-2204-arm` is not a valid label?

Maleware:
Outdated, closing this

PR #400, opened 2023-06-30
Rework Hadoop Dockerfile

lfrancke:
- It now uses a Builder step for the intermediate bits
- It builds FUSE
- A new script to download a new version of Hadoop, validate signatures etc. and uploads it again

### DIFF
--- a/hadoop/CHANGELOG.md
+++ b/hadoop/CHANGELOG.md
@@ -10,6 +10,12 @@
 [#347]: https://github.com/stackabletech/docker-images/pull/347
 [#400]: https://github.com/stackabletech/docker-images/pull/400
 
+### Changed
+
+- Upgraded to JMX Exporter 0.19 ([#400])
+
+[#400]: https://github.com/stackabletech/docker-images/pull/400
+
 ## [3.2.2-stackable0.7.0] [3.3.1-stackable0.7.0] [3.3.3-stackable0.3.0] [3.3.4-stackable0.3.0] - 2022-12-27
 
 ### Changed
@@ -18,7 +24,5 @@
   contains a layer which provides Vector. The creation of the stackable user
   and group happens in the stackable-base layer and is therefore removed from
   this image ([#285]).
-- Upgraded to JMX Exporter 0.19 ([#400])
 
 [#285]: https://github.com/stackabletech/docker-images/pull/285
-[#400]: https://github.com/stackabletech/docker-images/pull/400
--- a/hadoop/CHANGELOG.md
+++ b/hadoop/CHANGELOG.md
@@ -7,13 +7,11 @@
 - Add krb5-workstation and openssl packages - needed for Kerberos support ([#347]).
 - Add all necessary components to the image to mount HDFS using FUSE ([#400])
 
-[#347]: https://github.com/stackabletech/docker-images/pull/347
-[#400]: https://github.com/stackabletech/docker-images/pull/400
-
 ### Changed
 
 - Upgraded to JMX Exporter 0.19 ([#400])
 
+[#347]: https://github.com/stackabletech/docker-images/pull/347
 [#400]: https://github.com/stackabletech/docker-images/pull/400
 
 ## [3.2.2-stackable0.7.0] [3.3.1-stackable0.7.0] [3.3.3-stackable0.3.0] [3.3.4-stackable0.3.0] - 2022-12-27
--- a/hadoop/CHANGELOG.md
+++ b/hadoop/CHANGELOG.md
@@ -5,8 +5,10 @@
 ### Added
 
 - Add krb5-workstation and openssl packages - needed for Kerberos support ([#347]).
+- Add all necessary components to the image to mount HDFS using FUSE ([#400])
 
 [#347]: https://github.com/stackabletech/docker-images/pull/347
+[#400]: https://github.com/stackabletech/docker-images/pull/400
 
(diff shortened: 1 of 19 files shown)


lfrancke:
Once this is merged I'll also do a documentation PR

hadoop/Dockerfile
```diff
@@ -1,36 +1,41 @@
...
+RUN curl --fail -L https://repo.stackable.tech/repository/packages/hadoop/hadoop-${PRODUCT}-src.tar.gz | tar -xzC . && \
+    curl --fail -L https://repo.stackable.tech/repository/packages/hadoop/hadoop-${PRODUCT}.tar.gz | tar -xzC . && \
     ln -s /stackable/hadoop-${PRODUCT} /stackable/hadoop && \
-    ln -sf /stackable/bin/start-namenode /stackable/hadoop/bin && \
-    curl https://repo.stackable.tech/repository/packages/jmx-exporter/jmx_prometheus_javaagent-0.16.1.jar \
-    -o /stackable/jmx/jmx_prometheus_javaagent-0.16.1.jar && \
-    chmod -x /stackable/jmx/jmx_prometheus_javaagent-0.16.1.jar
+    rm -rf /stackable/hadoop/lib/native/examples && \
+    rm -rf /stackable/hadoop/share/doc && \
+    curl --fail https://repo.stackable.tech/repository/packages/jmx-exporter/jmx_prometheus_javaagent-${JMX_EXPORTER_VERSION}.jar -o /stackable/jmx/jmx_prometheus_javaagent-${JMX_EXPORTER_VERSION}.jar && \
```
maltesander, line 37: Can we do a link like this? That means we can stop hardcoding the version in the operators.
https://github.com/stackabletech/docker-images/blob/109af74441b0ba6ad7e784d020903e0e1bbd1c12/trino/Dockerfile#L35C1-L38C114

maltesander (commented):
I think this qualifies for a changelog entry?

hadoop/Dockerfile
lfrancke: Good idea. I added it.

Trino calls it `JMX_EXPORTER` while I call it `JMX_EXPORTER_VERSION`
Any preferences?

lfrancke: Do I then also need to change the operator?

lfrancke: Answer: Yes. I found it and will add it.

maltesander: > Good idea. I added it.
> 
> Trino calls it `JMX_EXPORTER` while I call it `JMX_EXPORTER_VERSION` Any preferences?

I like JMX_EXPORTER_VERSION better.

> Do I then also need to change the operator?

This would be somewhat breaking. It will only work with new images, not old ones. 
I just wanted to have it in every operator and migrate at some point.

hadoop/Dockerfile
```diff
@@ -1,36 +1,51 @@
...
+# And then we can also remove the symlink from this Dockerfile.
+RUN curl --fail -L "https://repo.stackable.tech/repository/packages/hadoop/hadoop-${PRODUCT}-src.tar.gz" | tar -xzC . && \
+    curl --fail -L "https://repo.stackable.tech/repository/packages/hadoop/hadoop-${PRODUCT}.tar.gz" | tar -xzC . && \
+    ln -s "/stackable/hadoop-${PRODUCT}" /stackable/hadoop && \
+    rm -rf /stackable/hadoop/lib/native/examples && \
+    rm -rf /stackable/hadoop/share/doc && \
+    curl --fail "https://repo.stackable.tech/repository/packages/jmx-exporter/jmx_prometheus_javaagent-${JMX_EXPORTER_VERSION}.jar" -o "/stackable/jmx/jmx_prometheus_javaagent-${JMX_EXPORTER_VERSION}.jar" && \
+    chmod -x "/stackable/jmx/jmx_prometheus_javaagent-${JMX_EXPORTER_VERSION}.jar" && \
+    ln -s "/stackable/jmx/jmx_prometheus_javaagent-${JMX_EXPORTER_VERSION}.jar" /stackable/jmx/jmx_prometheus_javaagent.jar && \
+    ln -s /stackable/jmx/jmx_prometheus_javaagent.jar /stackable/jmx/jmx_prometheus_javaagent-0.16.1.jar
```
soenkeliebau, line 47: I may be missing the point, but this still says 0.16.1?

hadoop/Dockerfile
lfrancke: I opted to use `JMX_EXPORTER`.
It looks better in conf.py and it is already used in Trino. I wanted to stay consistent.

lfrancke:
Can someone help me with the markdownlint issue?

hadoop/CHANGELOG.md
```diff
@@ -5,8 +5,16 @@
...
+- Add all necessary components to the image to mount HDFS using FUSE ([#400])
 
 [#347]: https://github.com/stackabletech/docker-images/pull/347
+[#400]: https://github.com/stackabletech/docker-images/pull/400
+
+### Changed
+
+- Upgraded to JMX Exporter 0.19 ([#400])
+
+[#400]: https://github.com/stackabletech/docker-images/pull/400
```
maltesander, line 17: We usually do it like this:
```suggestion
- Add all necessary components to the image to mount HDFS using FUSE ([#400])

### Changed

- Upgraded to JMX Exporter 0.19 ([#400])

[#347]: https://github.com/stackabletech/docker-images/pull/347
[#400]: https://github.com/stackabletech/docker-images/pull/400
```

PR #413, opened 2023-07-17
feat(airflow): Add s3fs python package, needed for pandas S3 access

sbernauer:
# Description

```[tasklist]
```

This enables us to e.g.
```
    @task
    def write_repos_to_s3(df: pandas.DataFrame):
        staging_table_name =''.join(random.choice(string.ascii_lowercase) for i in range(32))
        df.to_parquet(
            f"s3://{S3_BUCKET}/staging/github/{staging_table_name}/repos.parquet",
            storage_options={
                "key": S3_ACCESS_KEY_ID,
                "secret": S3_SECRET_ACCESS_KEY,
                "client_kwargs": {'endpoint_url': S3_ENDPOINT}
            }
        )
        return staging_table_name
```

### DIFF
--- a/airflow/Dockerfile
+++ b/airflow/Dockerfile
@@ -26,7 +26,8 @@ RUN microdnf update \
         && python3 -m venv --system-site-packages /stackable/app \
         && source /stackable/app/bin/activate \
         && pip install --no-cache-dir --upgrade pip \
-        && pip install --no-cache-dir apache-airflow[${AIRFLOW_EXTRAS}]==${PRODUCT} --constraint /tmp/constraints.txt
+        && pip install --no-cache-dir apache-airflow[${AIRFLOW_EXTRAS}]==${PRODUCT} --constraint /tmp/constraints.txt \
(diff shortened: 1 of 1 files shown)


lfrancke:
Also see here 
https://github.com/stackabletech/docker-images/pull/411/files :)

sbernauer:
I was thinking about the same, but I did not want to clog up the Dockerfile, so I put the reason in the commit message ;)
Added it to Dockerfile as well

sbernauer:
Hmm, if think technically I can't check the other boxes. Do we want to make some exceptions for pip? Is there any guide regarding the boms yet?

lfrancke (approved):

lfrancke:
No SBOM guide yet. I'm working on that.

So, this will need to change again but good enough for now

lfrancke:
As always. Those are just templates. Leave the boxes that make sense, remove the others.

I think this is good to go. I'm only on my phone but I believe I approved it as well

sbernauer:
Yeah you did, thanks!

PR #427, opened 2023-07-26
Mirror image action

dervoeti:
A Github Action to manually mirror an image in our OCI registry. It also signs it automatically.

### DIFF
--- a/.github/workflows/mirror.yaml
+++ b/.github/workflows/mirror.yaml
@@ -1,0 +1,45 @@
+name: Mirror container image
+on:
+  workflow_dispatch:
+    inputs:
+      src:
+        description: 'Location of the source image (example: registry.k8s.io/git-sync/git-sync:v3.6.4)'
+        required: true
+      dst:
+        description: 'Name and tag of the destination image (example: git-sync:v3.6.4)'
+        required: true
+
+env:
+  DST_REGISTRY: oci.stackable.tech
+  DST_PROJECT: mirror
+
+jobs:
+  mirror-image:
+    runs-on: ubuntu-latest
+    permissions:
+      id-token: write
+    steps:
+      - name: Login to Docker Hub
+        uses: docker/login-action@v2
+        with:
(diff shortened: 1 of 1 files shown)


.github/workflows/mirror.yaml
```diff
@@ -0,0 +1,46 @@
...
+  DST_REGISTRY: oci.stackable.tech
+  DST_PROJECT: mirror
+
+jobs:
+  mirror-image:
+    runs-on: ubuntu-latest
+    permissions:
+      id-token: write
+    steps:
+      - name: Login to Docker Hub
```
lfrancke, line 23:
```suggestion
      - name: Login to destination registry
```

lfrancke: Or `Login to oci.stackable.tech` or a similar name

lfrancke (changes requested):
Minor comment, otherwise looking good.

We talked about this and there's no easy way to verify source signatures generically, right?

dervoeti:
I think it would be possible, but the UI gets a bit complicated then. To verify the signature, the person triggering this workflow would have to somehow specify *how* the signature should be verified.
So there would be three options:
 - Image was signed with a private key, provide a public key for verification
 - Image was signed keyless, provide an issuer and a subject for verification
 - Image was not signed at all or the user does not want to verify it, then nothing needs to be provided

Alternatively, the user could specify a digest for the source image and verify that one manually using cosign. For the git-sync image in the example this would be:

`cosign verify registry.k8s.io/git-sync/git-sync@sha256:a794346111b84f7e3b9d65c63f1935b09c474b6cb6ca7486fe2cf97225c48620 --certificate-identity krel-trust@k8s-releng-prod.iam.gserviceaccount.com --certificate-oidc-issuer https://accounts.google.com`

(and then passing `registry.k8s.io/git-sync/git-sync@sha256:a794346111b84f7e3b9d65c63f1935b09c474b6cb6ca7486fe2cf97225c48620` as source image).

lfrancke (approved):

PR #432, opened 2023-08-02
Add "--fail" to curl commands

lfrancke:
# Description

This should fail the build if a download fails


adwk67 (approved):
lgtm!

PR #433, opened 2023-08-03
Remove Java DNS security props.

razvan:
# Description

Part of https://github.com/stackabletech/issues/issues/359

**NOTE:** This PR removes the properties that we appended to the security file. This means that the default properties remain unchanged (currently 10 secs for the negative cache and 30 secs for the positive one)

```[tasklist]
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,7 +18,7 @@ All notable changes to this project will be documented in this file.
 
 - Remove unused environment variable `AIRFLOW_UID` from the Airflow
   image ([#429]).
-- Java base image: hard-coded security properties for DNS caching ([#433])
+- Java base image: Remove hard-coded JVM security properties containing DNS cache settings. Going forward operators will configure DNS cache settings ([#433])
 
 [#419]: https://github.com/stackabletech/docker-images/pull/419
 [#429]: https://github.com/stackabletech/docker-images/pull/429
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,9 +18,11 @@ All notable changes to this project will be documented in this file.
 
 - Remove unused environment variable `AIRFLOW_UID` from the Airflow
   image ([#429]).
+- Java base image: hard-coded security properties for DNS caching ([#433])
 
 [#419]: https://github.com/stackabletech/docker-images/pull/419
 [#429]: https://github.com/stackabletech/docker-images/pull/429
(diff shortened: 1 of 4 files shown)


CHANGELOG.md
```diff
@@ -18,9 +18,11 @@ All notable changes to this project will be documented in this file.
 
 - Remove unused environment variable `AIRFLOW_UID` from the Airflow
   image ([#429]).
+- Java base image: hard-coded security properties for DNS caching ([#433])
```
sbernauer, line 21: Do we want to have a dedicated CHANGELOG for the java base image?
```suggestion
- Java base image: Re-enable DNS caching ([#433])
```

sbernauer (commented):
I don't disagree with this change, but we need to be aware caching was disabled because it caused problems ;) So this needs extensive testing I guess...

CHANGELOG.md
razvan: IMO no. The current situation shows that multiple changelogs don't work. I personally want a single changelog that is maintained - especially with the release entries.

Also I think my entry above is more accurate than your suggestion because the plan is to delegate these settings to the operators - which might decide to keep the cache disabled.

sbernauer: Agreed. Wasn't aware of the operator setting the dns settings in the future.
WDYT of
`- Java base image: Remove hard-coded JVM security properties containing DNS caching settings. Going forward operators will configure DNS caching settings ([#433])`

lfrancke:
I agree with Sebastian. Especially anything around ZooKeeper (e.g. HBase)

And we may want to set some limits instead of removing them all together?

sbernauer:
I think the plan is that the operators disable the cache when needed, which is a good idea, as in the end the operator should decide this, not the container image...
The only fear I have merging this right now is the time period it takes to properly implement this in all the product operators that need to have caching disabled. For the intermediate time being we would ship buggy products. This is especially bad if we wont get it done before 22.11 ;)

sbernauer:
But I'm also happy to approve this and let @razvan coordinate as he knows best on what to do in which order :)
I only have the concern mentioned above

razvan:
This is not meant to be merged _now_. I updated the description to make it clear.

CHANGELOG.md
razvan: Done https://github.com/stackabletech/docker-images/pull/433/commits/017e0ed4d2bd56891d836c5e4da0e03e7eccede4

PR #441, opened 2023-08-30
Rework Druid image to use a builder step and apply all around best practices

lfrancke:
# Description

This is similar to the recent PRs for Hadoop and Airflow
- It splits up building in two steps which saves a bit of space in the final image and makes the Dockerfile more readable (don't care about number of layers in intermediate builder)
- It adds `--fail` to all curl commands to fail if a download fails
- Sets environment variables

```[tasklist]
```


razvan:
Can this be closed? I'm trying to clean up stale branches

lfrancke:
Almost certainly! Thanks

PR #481, opened 2023-10-16
feat: parallel matrix build on product versions

razvan:


nightkr:
Where do these new arguments come from, https://github.com/stackabletech/image-tools/pull/1? Why was this split out into yet another repo? :/

nightkr:
And if image-tools *does* have to be a separate repo... at the very least, *please* pin the version used. Reproducibility is already enough of a mess as it is.

razvan:
Yes. This pr depends on https://github.com/stackabletech/image-tools/pull/1 and I will make a proper release of `image_tools` when that is approved.

soenkeliebau:
Are we planning to use image-tools separate from this repo anywhere? Otherwise folding it into this repo might make things easier on us indeed..

razvan:
Yes, the plan is to use it for operator images as well - for running preflight checks.

The code to run all these tools, `buildx`, `preflight` and  who knows what else we'll add in the future is already pretty complicated and needs proper testing and qa. This is a lot easier and faster if it's decoupled from the actual image building process.

maltesander (approved):
LGTM!

PR #484, opened 2023-10-16
disable signing

razvan:
- Disable dev image signing for testing


lfrancke:
Why?

lfrancke:
Sorry, I removed it from the queue again.
I have seen too much commented out code with no explanation in my life.

I'd like to understand the plan (even if it's just temporary please leave a comment either in code or in the PR)

razvan:
this is a test. signing code doesn't cope with the fact that not all runners produce images now.

we need to understand why jobs are canceled

PR #499, opened 2023-11-09
fix: add tzdata to superset image

adwk67:
# Description

- Redhat has removed `tzdata` from the ubi-minimal images. See [here](https://www.reddit.com/r/redhat/comments/12maj9w/ubi_feedback/) and [here](https://bugzilla.redhat.com/show_bug.cgi?id=2223028). 
- Superset relies on [ZoneInfo](https://docs.python.org/3/library/zoneinfo.html#data-sources) to resolve time zones, and this is done by searching first under `TZPATH` (which is empty due to the point above) or for the tzdata python package.
- That package is added in this PR (airflow already has tzdata in its list of dependencies). 

```[tasklist]
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -24,6 +24,7 @@ All notable changes to this project will be documented in this file.
 - opa: add new version upload script ([#471]).
 - spark: added versions 3.4.1, 3.5.0 ([#475]).
 - superset: add new version 2.1.1, 3.0.1 ([#482], [#489]).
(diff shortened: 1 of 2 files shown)


sbernauer:
Thanks for the catch and the nice description!
Could you please add the description as comment above the package, so that we keep this valuable information?

adwk67:
> Thanks for the catch and the nice description! Could you please add the description as comment above the package, so that we keep this valuable information?

Done!

NickLarsenNZ (approved):
LGTM

PR #509, opened 2023-11-23
Add topology provider jar to Hadoop image.

soenkeliebau:
# Description

This adds an extra jar file to the Hadoop image which contains an implementation of the `org.apache.hadoop.net.DNSToSwitchMapping` interface to generate rack IDs based on Kubernetes labels, so that HDFS knows the underlying Kubernetes architecture and can distribute blocks properly.

Related to https://github.com/stackabletech/hdfs-operator/issues/325

```[tasklist]
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,7 @@ All notable changes to this project will be documented in this file.
 
 ### Added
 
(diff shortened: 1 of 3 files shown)


adwk67 (dismissed):
lgtm (testing to come with the operator changes). Just one nit: do we want to change line 147 to use curl (with --fail) in case the jar-fetch fails for some reason?

PR #511, opened 2023-11-24
Added note on file origin

razvan:


hive/stackable/bin/start-metastore
```diff
@@ -1,5 +1,7 @@
 #!/usr/bin/env bash
 #
+# This file was created by the Stackable developers.
```
sbernauer, line 3: Suggestion to remove copy/pasted content
```
# This file was created by the Stackable developers.
#
# Usage: start-metastore <options>
# 
# Checks if the metastore database schema is initialized. If so it starts the metastore,
# otherwise it tries to initialize the schma first.
#
```

sbernauer (changes requested):

hive/stackable/bin/start-metastore
razvan: I don't understand

maltesander: The usage description contains namenode etc. 
```
# Usage: start-metastore <options> namenode [namenode-options]
```

razvan: I see. Done.

sbernauer (approved):
Thanks!

PR #513, opened 2023-11-28
feat(testing-tools): add keycloak

NickLarsenNZ:
# Description

Add Keycloak so that we don't need to download them in the demo and kuttl tests.

```[tasklist]
```

### DIFF
--- a/testing-tools/Dockerfile
+++ b/testing-tools/Dockerfile
@@ -2,6 +2,7 @@ FROM python:3.11-slim-bullseye@sha256:286f2f1d6f2f730a44108656afb04b131504b610a6
 
 ARG PRODUCT
 ARG RELEASE
+ARG KEYCLOAK_VERSION
 
 LABEL name="Stackable Testing Tools" \
       maintainer="info@stackable.tech" \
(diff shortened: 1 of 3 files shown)


testing-tools/Dockerfile
```diff
@@ -43,6 +44,11 @@ RUN apt-get update && \
 COPY testing-tools/python /stackable/python
 COPY testing-tools/licenses /licenses
 
+ENV PATH=/stackable/keycloak/bin:$PATH
+# todo: do we need to maintain this via the mirror script?
+RUN curl --fail https://repo.stackable.tech/repository/packages/keycloak/keycloak-${KEYCLOAK_VERSION}.tar.gz | tar -xzC /stackable && \
```
NickLarsenNZ, line 49: @lfrancke we have checked the sha1 hash and uploaded [keycloak-23.0.0.tar.gz](https://github.com/keycloak/keycloak/releases/download/23.0.0/keycloak-23.0.0.tar.gz) manually.

Should we add a script to take care of this like what it done for [hbase](https://github.com/stackabletech/docker-images/blob/89e39e9db8e06846dc825076c2a74c55210298a6/hbase/upload_new_hbase-operator-tools_version.sh)?

testing-tools/Dockerfile
```diff
@@ -43,6 +44,11 @@ RUN apt-get update && \
 COPY testing-tools/python /stackable/python
 COPY testing-tools/licenses /licenses
 
+ENV PATH=/stackable/keycloak/bin:$PATH
+# todo: do we need to maintain this via the mirror script?
```
sbernauer, line 48:
```suggestion
```

sbernauer (dismissed):
LGTM otherwise, thanks!

testing-tools/Dockerfile
NickLarsenNZ: I think we'll leave it for now and it can be done when the Keycloak dependency is updated.

testing-tools/Dockerfile
NickLarsenNZ: Done in 951720f6fe42ee4a37414d9d164ce61888e93cd1

sbernauer (approved):

PR #516, opened 2023-12-05
Changes default async-profiler support version from 1.x to 2.x

lfrancke:
# Description

https://issues.apache.org/jira/browse/HBASE-25685

See this comment:

> Here is suggestion... here for branch-2.3/2.4 #3078... i.e. presume asyncprofiler1 but if asyncprofiler2 is there, allow setting output=html. For branch-2 and master, presume asyncprofiler2... thats #3079

This changes our HBase to presume asyncprofiler 2 and thus makes it work without extra parameters.

```[tasklist]
```


nightkr:
Is this still relevant? The upstream issue claims that this was fixed in 2.4.3 and 2.5.x, which are both older than we support (..then again, it's also older than the 2.4.12 that this PR applies it to).

lfrancke:
I have not forgotten about this. I just need to find time to look into it again :(

PR #519, opened 2023-12-11
feat: make preflight a standalone manual workflow

razvan:


conf.py
```diff
@@ -374,7 +374,7 @@
     "druid": {"id": "626140028ccb9938ba3cfde7"},
     "hadoop": {"id": "6261407f887d6e0b8614660c"},
     "hbase": {"id": "62614109992bac3f9a4a24b8"},
-    "hive": {"id": "626140028ccb9938ba3cfde7"},
+    "hive": {"id": "626140806812078a392dceaa"},
```
sbernauer, line 377: I have no idea what this is about but I'm sure you do :)

sbernauer (approved):

conf.py
razvan: Yes, the id was wrong :)

PR #526, opened 2023-12-18
Build Hadoop from source

### LINKED ISSUE #525: hadoop: Build Hadoop from source to be able to apply patches
Hadoop should be entirely built from source, so that patches can be applied. Currently, only the HDFS native client is built to create the fuse-dfs executable.

Instead of building it in the java-base image, it could be easier to use the Dockerfile provided in the Hadoop sources (https://github.com/apache/hadoop/blob/rel/release-3.3.6/dev-support/docker/Dockerfile). One reason is that the libtirpc-devel package is required which is not available in the used repositories.

siegfriedweber:
# Description

Build Hadoop from source

Closes #525 

```[tasklist]
```


siegfriedweber:
https://ci.stackable.tech/view/02%20Operator%20Tests%20(custom)/job/hdfs-operator-it-custom/110/

lfrancke (approved):
Thank you!
I looked at it and it looks good.
I'm a bit surprised by the different protobuf versions as I thought I already looked at that but I must have made a mistake, thanks for catching that.

I built the image locally and the build worked. I have not tested it though.

PR #537, opened 2024-01-25
feat(java-base): Add openjdk-devel package

sbernauer:
# Description

As decided in Arch meeting on 2024-01-25
For https://github.com/stackabletech/spark-k8s-operator/discussions/340

```[tasklist]
```


java-base/Dockerfile
```diff
@@ -19,6 +19,9 @@ RUN microdnf update && \
     microdnf install \
     # Obviously needed to run Java programs
     java-${PRODUCT}-openjdk-headless \
+    # Includes tools such as jps or jmap (which is e.g. needed by the Spark UI, see https://github.com/stackabletech/spark-k8s-operator/discussions/340)
+    # We consider them usefull enough to justify putting them in all Java images
```
lfrancke, line 23:
```suggestion
    # We consider them useful enough to justify putting them in all Java images
```

sbernauer: Where is the spellcheck action when we need it the most? :D

lfrancke:
This should probably also change the other Dockerfiles which might manually install devel right now including the parameters we added for that

Examples: 
https://github.com/stackabletech/docker-images/blob/main/hadoop/Dockerfile#L17
https://github.com/stackabletech/docker-images/blob/main/conf.py#L63-L64

sbernauer:
Good catch, wasn't ware of them!

java-base/Dockerfile
lfrancke: :)

lfrancke (approved):

PR #546, opened 2024-02-06
Ubi8-rust-builder multi-arch workflow using BuildJet for ARM64

Maleware:
# Description

part of

This PR is solving one part of  https://github.com/stackabletech/issues/issues/475 and opt to solve the following:

Multi architecture manifest list (aarch64 and X86_64) for the ubi8-rust-builder.

Therefore this extends the Makefile to create and publish both architectures. The workflow got extended to a [BuildJet](https://buildjet.com/) - building instance. 

both Archs get build, in a second step both images get pulled, bundeld in a manifest list and this list get's pushed to Nexus. 

This is enabling us to build multi architecture operator images out of the box for every user. 

```[tasklist]
```


lfrancke:
If it is still WIP, can it be converted do a draft?

.github/workflows/docker.yml
```diff
@@ -10,12 +10,26 @@ on:
 
 jobs:
   build:
-    name: Build and push Docker image
-    runs-on: ubuntu-latest
+    name: Build and publish both arch
```
nightkr, line 13: Each individual subtask here still only applies to one arch

Makefile
```diff
@@ -2,13 +2,33 @@
...
 endef
 
 define build
-	@docker build --force-rm -t "${REPO}/${1}:${TAG}" -t "${REPO}/${1}:latest" -f $(1)/Dockerfile .
+	@docker build --force-rm -t "${REPO}/${1}:${TAG}-${ARCH}" -t "${REPO}/${1}:latest-${ARCH}" -f $(1)/Dockerfile .
+endef
+
+# Tag for arm64 is fixed since this will run on gh ubuntu-latest
+define manifest
+	@docker manifest create "${REPO}/${1}:latest" ${REPO}/${1}:latest-${ARCH} ${REPO}/${1}:latest-aarch64
```
nightkr, line 17: Won't this break on ARM since it only links in aarch64 twice?

nightkr, line 17: Don't love the racy way that this works... Isn't there a way for us to save the build hashes between the steps?

Makefile
```diff
@@ -2,13 +2,33 @@
...
+endef
+
+# Pulling both images after building them, ugly
+# TODO: find a better way
+define pull-arm64
+	@docker pull ${REPO}/${1}:latest-aarch64 
+endef
+
+define pull-amd64
+	@docker pull ${REPO}/${1}:latest-${ARCH}
```
nightkr, line 31: Same issue relying on `${ARCH}` here.

Makefile
```diff
@@ -19,6 +39,19 @@ push-ubi8-rust-builder: NAME = ubi8-rust-builder
...
+	$(call pull-arm64,${NAME})
+	$(call pull-amd64,${NAME})
+
+build-manifest-list: NAME = ubi8-rust-builder
+build-manifest-list:
+	$(call manifest,${NAME})
+
+push-manifest-list: NAME = ubi8-rust-builder
+push-manifest-list: login pull-ubi8-rust-builder build-manifest-list 
+	$(call manifest_push,${NAME})
```
nightkr, line 53: I assume this is meant to deduplicate `NAME`, yet we have to list it separately for each target anyway?

Makefile
Maleware: Yeah it would, but I assume this only is meant to be for the build-pipeline. The create manifest list is meant to only run on ubuntu-latest, therefore I went down the easy path.

Not sufficient?

Maleware: Why a racy way?

Makefile
Maleware: See comment above

Makefile
nightkr: I still don't see a reason to not just hard-code amd64 (like you already do for arm64).

Maleware: Ah, yeah we could do that. No pains with that

Makefile
Maleware: I don't get it.... The targets should be generated via `latest-{ARCH}` but probably I misunderstand the issue.

Makefile
nightkr: It assumes that nothing else writes to `latest-$arch` in between the steps.

Maleware: Ah yeah. The way the pipeline is designed now (manifest_create needs both builds to succeed and is then fairly fast) should`TM` prevent this. 

I can look for a maybe more save way to amend with sha's rather than tags

Makefile
Maleware: Removed pull, unnecessary. Hardcoded arch's wherever possible

Makefile
Maleware: Hardcoded arch wherever possible

.github/workflows/docker.yml
Maleware: done

Makefile
Maleware: @nightkr so yeah, technically that's possible... but hacky.

As much as I can come up with we would do something like this:
```
ARCH=$(arch) &&
          sha-$ARCH=$(docker images --no-trunc --quiet docker.stackable.tech/stackable/ubi8-rust-builder:latest-$ARCH) \
          echo "sha-$ARCH=$sha-$ARCH" >> $GITHUB-OUTPUT
```

and teach the `Makefile` to use those sha's rather then the tags.

I don't really like the solution, as we could mitigate the race condition with this condition:
```
  create_manifest:
    name: Create and publish manifest list
    runs-on: ubuntu-latest
    needs: ["build"]
```

WDYT?

Makefile
nightkr: Is there a reason we don't define `NAME` on the top level, like `REPO` and co?

Makefile
Maleware: Yes, that would be a rework in the script. I'm not sure if I like that more than having it obvious in the pipeline tbh....

Makefile
Maleware: not really, I just copied the other steps. i can do that :)

nightkr (approved):

PR #549, opened 2024-02-08
Use mirrored inotify-tools and use Vector 0.35.0 in all images

siegfriedweber:
# Description

Use the inotify-tools from the mirror and use Vector 0.35.0 in all images

EPEL 8 was not available what prevented us from building images.

```[tasklist]
```


Maleware (approved):
LGTM!

PR #566, opened 2024-02-16
feat(hadoop): Add hdfs-utils

sbernauer:
# Description

Depends on https://github.com/stackabletech/hdfs-group-mapper/pull/2

```[tasklist]
```

### DIFF
--- a/conf.py
+++ b/conf.py
@@ -84,7 +84,7 @@ products = [
                 "async_profiler": "2.9",
                 "jmx_exporter": "0.20.0",
                 "protobuf": "2.5.0",
-                # no HDFS utils, as e.g. patches from https://github.com/apache/hadoop/pull/6553 are missing. 3.2.x will probably removed shortly anyway.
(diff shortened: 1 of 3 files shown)


lfrancke:
What are the hdfs-utils?

conf.py
```diff
@@ -84,6 +84,7 @@
                 "async_profiler": "2.9",
                 "jmx_exporter": "0.20.0",
                 "protobuf": "2.5.0",
+                # no HDFS utils, as e.g. patches from https://github.com/apache/hadoop/pull/6553 are missing. 3.2.x will probably removed shortly anyway.
```
adwk67, line 87:
```suggestion
                # no HDFS utils, as e.g. patches from https://github.com/apache/hadoop/pull/6553 are missing. 3.2.x will probably removed in version 24.3 anyway.
```

adwk67 (approved):
lgtm. Just one nit, which i'll leave up to you (so that the comment about the removal still makes sense in the distant future!)

conf.py
sbernauer: Let's hope for it :)

adwk67 (approved):

PR #582, opened 2024-03-02
Adds convenience functions to the base image

lfrancke:
# Description

- Adds iputils (for ping), less, nano and tar
- Adds "ll" and ".." as aliases, enables color for "ls"
- Adds a shell prompt which includes the user, hostname and current path

```[tasklist]
```


stackable-base/Dockerfile
```diff
@@ -16,11 +24,33 @@ ARG PRODUCT
 COPY stackable-base/stackable/dnf.conf /etc/dnf/dnf.conf
 
 RUN microdnf update && \
-    microdnf --assumeyes install shadow-utils && \
+    microdnf --assumeyes install \
```
siegfriedweber, line 27: `tar` could be added because it is required for `kubectl cp`.

https://kubernetes.io/docs/reference/kubectl/generated/kubectl_cp/:
> Requires that the 'tar' binary is present in your container image. If 'tar' is not present, 'kubectl cp' will fail.

lfrancke:
Whoever reviews this: Please don't take this as a given. If you don't like any of these changes bring it up please. Consider this PR a suggestion.

airflow/Dockerfile
```diff
@@ -13,9 +13,6 @@ ARG STATSD_EXPORTER
 ARG TARGETARCH
 ARG TARGETOS
 
-# https://github.com/hadolint/hadolint/wiki/DL4006
-SHELL ["/bin/bash", "-o", "pipefail", "-c"]
-
```
NickLarsenNZ, line 18: **question**: why did the shell get unset in some images?

kafka/Dockerfile
```diff
@@ -3,8 +3,18 @@ FROM stackable/image/java-base AS builder
...
+    cyrus-sasl-devel \
+    gcc-c++ \
+    libcurl-devel \
+    make \
+    openssl-deve \
+    tar \
+    wget \
+    which \
+    zlib \
+    zlib-devel && \
```
NickLarsenNZ, line 17: **praise**: I much prefer the multiline separation :smile:

stackable-base/Dockerfile
```diff
@@ -5,6 +5,14 @@ FROM registry.access.redhat.com/ubi8/ubi-minimal:8.9@sha256:87bcbfedfd70e67aab38
 # intentionally unused
 ARG PRODUCT
 
+# Sets the default shell to Bash with strict error handling and robust pipeline processing.
+# "-e": Exits immediately if a command exits with a non-zero status
+# "-u": Treats unset variables as an error, preventing unexpected behavior from undefined variables.
+# "-o pipefail": Causes a pipeline to return the exit status of the last command in the pipe that failed, ensuring errors in any part of a pipeline are not ignored.
+# "-c": Allows the execution of commands passed as a string
+# This is automatically inherited in all other Dockerfiles that use this unless it is overwritten
+SHELL ["/bin/bash", "-euo", "pipefail", "-c"]
```
NickLarsenNZ, line 14: **praise**: I also like this explanation of the flags/args.

NickLarsenNZ (commented):
Just left some comments (in the [conventional comment][1] style).

[1]: https://conventionalcomments.org/

airflow/Dockerfile
lfrancke: It should be unset in all images actually.

I can revert this but it was inconsistent between images and I looked and it seems as if it is inherited by the base images so we only have to define this once.

I don't feel strongly about this though. I can also leave it in and just make it consistent.

stackable-base/Dockerfile
lfrancke: Thanks. As we only have this three times or so now I decided to include it. And it also contains the sentence about the inheritance.

kafka/Dockerfile
lfrancke: Looking at this...isn't there a typo? openssl-deve -> openssl-devel?

NickLarsenNZ: Indeed there is. I didn't notice it (but it would have been picked up during the next build).

lfrancke: I'll take another look at it. Maybe I committed something unintentionally here as Kafka is the next thing on my list to look at. Maybe this was meant to go on another branch.

stackable-base/Dockerfile
siegfriedweber, line 14: [hadolint](https://github.com/hadolint/hadolint) complains about the missing SHELL command in the other Dockerfiles:

```
airflow/Dockerfile:46 DL4006 warning: Set the SHELL option -o pipefail before RUN with a pipe in it. If you are using /bin/sh in an alpine image or if your shell is symlinked to busybox then consider explicitly setting your SHELL to /bin/ash, or disable this check
```

This warning should be ignored by adding the following line to all Dockerfiles which use stackable-base:

```Dockerfile
# hadolint global ignore=DL4006
```

siegfriedweber (changes requested):

stackable-base/Dockerfile
lfrancke: Thanks! I'll add that.

@razvan is it too late to get this into 24.3?

lfrancke: This should be done.

Maleware (approved):
LGTM! Builded Kafka from scratch and looks great! Thanks.

lfrancke:
Thanks everyone! On to a bright new future for our images :)

PR #603, opened 2024-03-26
Fix/manifest list nexus ubi8

Maleware:
# Description

Fixing that ubi8 rust builder currently is not pushed to nexus.


Maleware:
Can be closed and omitted since stackable-operators pull ubi8-rust-builder from harbor with https://github.com/stackabletech/operator-templating/pull/352

PR #604, opened 2024-03-26
Fix: Using Docker Build & Push action and remove login from manifest list creation

Maleware:
# Description

Fix: Using Docker Build & Push action and remove login from manifest list creation


dervoeti (approved):
LGTM

PR #658, opened 2024-05-08
Build Phoenix from source

dervoeti:
# Description
Fixes https://github.com/stackabletech/docker-images/issues/660

We added a script to mirror the Phoenix sourcecode to Nexus and build Phoenix in the Dockerfile (process is the same as the one for hbase-operator-tools).

It's a rather small change, we experimented with different approaches to build Phoenix and settled on the one that's in line with how other tools are build from source in our images at the moment.

The [integration test for Phoenix](https://github.com/stackabletech/hbase-operator/blob/main/tests/templates/kuttl/smoke/60-test-phoenix.yaml) passed successfully, we also did some manual tests.

```[tasklist]
```

<details>

```shell
```

</details>


hbase/upload_new_phoenix_version.sh
```diff
@@ -0,0 +1,71 @@
...
+function cleanup {
+  rm -rf "$WORK_DIR"
+}
+
+# register the cleanup function to be called on the EXIT signal
+trap cleanup EXIT
+
+cd "$WORK_DIR" || exit
+
+src_file=phoenix-$VERSION-src.tar.gz
```
sbernauer, line 34: You could argue it does not cost us anything to also download the pre-build version and it could save us the day in the future (e.g. quickly pull it on your machine and test something). But that's 100% speculative ;)

sbernauer (approved):
LGTM!

hbase/upload_new_phoenix_version.sh
dervoeti: Yeah we'll likely improve this whole "build from source" workflow at some point in the future (how to apply custom patches, how to test upcoming versions, how to track where we pulled stuff from in a machine-readable way etc.). I'd say it's out of scope for now, the main goal was just about building Phoenix from source (and testing it) instead of pulling the binaries.

PR #664, opened 2024-05-10
Move kcat into its own image and reuse it in kafka-testing-tools and …

lfrancke:
# Description

Move kcat into its own image and reuse it in kafka-testing-tools and kafka

This also fixes the build failure of kafka-testing-tools on ARM.

```[tasklist]
```

<details>

```shell
```

</details>


kcat/Dockerfile
```diff
@@ -0,0 +1,40 @@
...
+# Ignoring DL3038 globally because set `assumeyes=True` in dnf.conf in our base image
+# Ignoring DL4006 globally because we inherit the SHELL from our base image
+# hadolint global ignore=DL3038,DL4006
+
+# Normally we would use stackable/image/stackable-base here, *but* we fail to link kcat at the end with error messages
+# shown in Snippet 1, which we were not able to solve.
+
+# Not tagging base image because it is built as part of the same process
+# hadolint ignore=DL3006
+FROM stackable/image/java-base AS builder
```
sbernauer, line 12: I think it's also a good thing to keep Snippet 1
```suggestion
# Normally we would use stackable/image/stackable-base here, *but* we fail to link kcat at the end with error messages
# shown in Snippet 1, which we were not able to solve.
# Not tagging base image because it is built as part of the same process
# hadolint ignore=DL3006
FROM stackable/image/java-base AS builder
```

sbernauer (changes requested):
LGTM otherwise

kcat/Dockerfile
lfrancke: Your comment is actually already in there. See above.
But I forgot to copy the actual snippet.

lfrancke: [`fd60ba7` (#664)](https://github.com/stackabletech/docker-images/pull/664/commits/fd60ba707e7f662bb5cb72a19ebf670ef2de3c40)

sbernauer: Oh, sorry didn't see it

sbernauer (approved):

PR #665, opened 2024-05-13
Add java-devel as base layer for builder stages

razvan:
# Description

Images build successfully: https://github.com/stackabletech/docker-images/actions/runs/9065581172

**Remove the java development tools from final images.**

Some notes on the implications:

* the `openjdk-headless` only installs a JRE
* the `openjdk-devel` package is needed to build products from source (duh!) but also for log4shell checks
* the `openjdk-devel` package brings in a lot of dependencies with known CVEs like (X11 libs)

```[tasklist]
```

<details>

```shell
```

</details>


sbernauer:
To be honest I'm not really sure what I should think of this.

* (+) It simplifies the development
* (+) It should decrease the build time locally
* (-) It increases the build time across the board in the CI (we currently already struggle with the few ARM runners)
* (-) I think in the past @lfrancke told me to reduce the number of packages not only in the final but also in the builder image. Not sure if this is still the case though

Back in the days I would have brought it up in the Arch meeting, WDYT of throwing this on the Decisions board instead?

razvan:
> Thanks for giving this a try! To be honest I'm not really sure what I should think of this.

Yes, I'm not sure what to make of it either. But if we want to exclude devel packages from the final images **and** have no  duplication, it's the only way,

A big improvement in performance would be if we used published images of `java-devel` and` java-base`.

> Back in the days I would have brought it up in the Arch meeting, WDYT of throwing this on the Decisions board instead?

ok

sbernauer:
AFAIK we need `java-xxx-openjdk-devel` to build Java programs.
After we (together) fiddled around with `java-xxx-openjdk-devel` I think this actually LGTM

NickLarsenNZ:
Only thing I can say is that build time vulnerabilities still have the potential for affecting the resulting artifact. So we will still need to deal with the devel CVEs in some way, but perhaps the scope is reduced?

nightkr:
Haven't reviewed the PR contents, but conceptually :+1:.

> (-) I think in the past @lfrancke told me to reduce the number of packages not only in the final but also in the builder image. Not sure if this is still the case though

Don't think this changes much? Still the same build-time deps, fewer runtime deps is a win.

> A big improvement in performance would be if we used published images of java-devel and java-base.

Really not keen on going back to "use whatever java-devel is published" (or worse, "use whatever java-devel you have cached locally"). It would be nice if Docker had a way to notice that our java-devel-to-be-built matches a prebuilt java-devel exactly (... like Nix), but that's not practical as long as Dockerfiles expect to be able to download anything they want from the internet.

> Only thing I can say is that build time vulnerabilities still have the potential to affect the resulting artifact. So we will still need to deal with the devel CVEs in some way, but perhaps the scope is reduced?

Yeah, but there's a pretty huge difference between "build process could potentially have been backdoored if we knew about a discovered-far-in-the-future vulnerability" and "can be exploited by anyone until patched". Especially if the build is reproducible so that we (and others) can determine in the future whether or not such a build-time exploit actually happened (... not that we are atm, but a girl can dream right?).

sbernauer (approved):
LGTM as discussed in meeting

sbernauer:
We had a meeting (with 8 people) and there was no big discussion, no one objected

PR #675, opened 2024-05-15
chore/CVE-2022-1471

xeniape:
# Description
Bump dependency versions which contain fixes for CVE-2022-1471

```[tasklist]
```

<details>

```shell
```

</details>


hbase/Dockerfile
```diff
@@ -25,14 +25,16 @@ COPY hbase/licenses /licenses
 USER stackable
 WORKDIR /stackable
 
+## copy patches
+COPY --chown=stackable:stackable hbase/stackable/patches /stackable/patches
```
nightkr, line 29: Is there a reason to bump this up above the `curl`? This means that the source code will be redownloaded every time the patches are modified.

xeniape: I thought it was more readable since the patches are not only used for hbase patches but phoenix as well, no other reason. Can be changed back as well :)

sbernauer: Ideally we always only copy the needed stuff just before we need it to cache as much stuff as possible.
So this means pulling in the hbase and phoenix patches just before build :)

hbase/stackable/patches/apply_patches.sh
```diff
@@ -4,14 +4,21 @@
...
-PATCH_DIR="patches/$VERSION"
+# Check if $2 (VERSION) is provided
+if [ -z "${2-}" ]; then
+  echo "Please provide a value for VERSION as the second argument."
+  exit 1
+fi
+
+PRODUCT="$1"
+VERSION="$2"
+PATCH_DIR="patches/$PRODUCT/$VERSION"
```
nightkr, line 21: I see what this is getting at. It would be nice to unify these scripts across all the products at some point, but I think it's fine to defer that to a separate issue.

hbase/stackable/patches/hbase/2.4.17/006-HBASE-jruby-version-bump.patch
```diff
@@ -0,0 +1,11 @@
+--- a/pom.xml	2023-03-31 20:03:27.000000000 +0200
++++ b/pom.xml	2024-05-14 15:52:58.901324926 +0200
+@@ -576,7 +576,7 @@
+     <wx.rs.api.version>2.1.1</wx.rs.api.version>
+     <glassfish.jsp.version>2.3.2</glassfish.jsp.version>
+     <glassfish.el.version>3.0.1-b08</glassfish.el.version>
+-    <jruby.version>9.2.13.0</jruby.version>
++    <jruby.version>9.4.7.0</jruby.version>
```
nightkr, line 8: Have we tested this? [According to Wikipedia](https://en.wikipedia.org/wiki/JRuby#Release_history), this means effectively upgrading from Ruby 2.5.7 to 3.1. Things may have changed, but AFAIK, Ruby's versioning is so loose that even patch level bumps are sometimes incompatible.

hbase/stackable/patches/phoenix/5.1.3/001-PHOENIX-omid-deps-version-bump.patch
```diff
@@ -0,0 +1,29 @@
...
++        <version>1.1.2</version>
+         <exclusions>
+           <exclusion>
+             <groupId>org.testng</groupId>
+@@ -1326,7 +1326,7 @@
+       <dependency>
+         <groupId>org.apache.omid</groupId>
+         <artifactId>omid-common</artifactId>
+-        <version>${omid.version}</version>
++        <version>1.1.2</version>
```
nightkr, line 26: Couldn't this just be a bump to `omid.version`, similarly to the JRuby patch above?

hbase/stackable/patches/apply_patches.sh
nightkr: #677

hbase/stackable/patches/hbase/2.4.17/006-HBASE-jruby-version-bump.patch
xeniape: I ran the integration tests in the hbase-operator, which at least succeeded. Are there any other ways to test this? I'm not familiar with Ruby or how it is used by HBase.

nightkr: I'm not familiar with where it's used either.. @lfrancke might have more of a clue?

Not sure about hbase, but most of our integration tests don't do a great job exercising the products beyond "does it launch? can I connect?". :/

hbase/stackable/patches/phoenix/5.1.3/001-PHOENIX-omid-deps-version-bump.patch
nightkr: Oh, yikes. I'd rather consider the held-back versions the exception, but that also raises followup questions like.. "is omid 1.1.2 actually compatible with omid-hbase-common-hbase2.x 1.1.1? why did they need to drop it here?".

(To be clear, I'm not complaining at you here, more just being cautious about what the implications would be...)

hbase/stackable/patches/hbase/2.4.17/006-HBASE-jruby-version-bump.patch
nightkr: Yeahh, that was always broken for some reason that I still haven't had the energy to investigate properly.

sbernauer: You could also give the [hbase-hdfs-load-cycling-data demo](https://docs.stackable.tech/home/stable/demos/hbase-hdfs-load-cycling-data) a try

hbase/stackable/patches/phoenix/5.1.3/001-PHOENIX-omid-deps-version-bump.patch
xeniape: Seems like in the next phoenix version (5.2.0) they replaced omid-hbase-common-hbase2.x with omid-hbase-common (I base this observation solely on the name), which seems to be compatible with omid 1.1.2, they updated these versions as well. https://repo1.maven.org/maven2/org/apache/phoenix/phoenix/5.2.0/phoenix-5.2.0.pom

So the vulnerability coming from phoenix should be fixed with a new phoenix version, no patch needed then?

hbase/stackable/patches/hbase/2.4.17/006-HBASE-jruby-version-bump.patch
lfrancke: JRuby is used in the HBase shell only as far as I know.
We might be using that in tests etc. so it's definitely worth checking whether that still works properly.

hbase/stackable/patches/phoenix/5.1.3/001-PHOENIX-omid-deps-version-bump.patch
nightkr: Honestly, I'd have to leave that question to someone more knowledgeable about Phoenix than me..

xeniape: After some more digging: phoenix versions 5.x require Hadoop3, which HBase 2.x does not built with out of the box. Since we are building from source we might change that but I'm not sure if that has some other implications or problems down the line (https://github.com/apache/phoenix/blob/master/BUILDING.md#hbase-2-and-hadoop-3)
So not sure if it would be more reasonable to update the phoenix version together with the HBase version.

hbase/Dockerfile
xeniape: Reverted any changes regarding the patches folder and its use in the Dockerfile since the increase of the phoenix version itself removes the need for any phoenix patches for now.

lfrancke (approved):
It looks good to me but I didn't run any tests.

sbernauer:
FYI, Trino 447 removed support for Phoenix 5.1 (see https://trino.io/docs/current/release/release-447.html#phoenix-connector)

PR #676, opened 2024-05-15
feat: OPA build from source

fhennig:
# Description

We tested that the smoke test runs :+1: 

```[tasklist]
```

<details>

```shell
```

</details>


opa/Dockerfile
```diff
@@ -58,13 +58,32 @@ WORKDIR /daemontools/admin/daemontools-${DAEMONTOOLS_VERSION}
...
 ARG PRODUCT
 ARG RELEASE
 ARG TARGETARCH
 ARG TARGETOS
 
+ENV GOARCH=$TARGETARCH
+ENV GOOS=$TARGETOS
+
+RUN microdnf update && \
+    microdnf install git go gzip tar
```
Maleware, line 72:
```suggestion
    microdnf install --assumeyes \
        git \
        go \
        gzip \
        tar && \
        microdnf clean all
```

Maleware: We use as convention one line per installed tool. I would personally appreciate a comment why / for what you install those tools.

opa/Dockerfile
```diff
@@ -58,13 +58,32 @@ WORKDIR /daemontools/admin/daemontools-${DAEMONTOOLS_VERSION}
 
 RUN package/install
 
-FROM stackable/image/vector
+FROM stackable/image/vector as opa-builder
```
Maleware, line 61: I think stackable-base is sufficient as a builder image at this point.

Maleware: You would need to add 
```
"stackable-base": "1.0.0",
```
to `conf.py` tested it locally and it was sufficient. you can also add stackable-base to opa-bundle-builder afaik, but @siegfriedweber might have a veto for that.

Maleware (changes requested):
Thanks for your work, just a few nitty things.

opa/Dockerfile
lfrancke: And could you add a hadolint ignore to this line? Look at Kafka for an example.

opa/Dockerfile
lfrancke: Yes, Maxi is right.
Please split it across multiple lines.
`--assumeyes` is not needed. That is now baked into the base image and also yes: Please add a comment for why everything is needed. Thanks!

opa/Dockerfile
```diff
@@ -58,13 +58,32 @@ WORKDIR /daemontools/admin/daemontools-${DAEMONTOOLS_VERSION}
...
+
+RUN curl --fail -L "https://repo.stackable.tech/repository/packages/opa/opa_${PRODUCT}.tar.gz" -o opa.tar.gz && \
+    tar -zxvf opa.tar.gz && \
+    mv opa-${PRODUCT} opa
+
+WORKDIR /opa
+
+RUN go build -o opa -buildmode=exe
+
+FROM stackable/image/vector
```
lfrancke, line 82: See comment above about the hadolint ignore

lfrancke (changes requested):

opa/Dockerfile
fhennig: Will do. Is `microdnf clean all` really needed? This is just a builder image, so it gets thrown away anyways

Maleware (approved):
Another Job well done :) Thanks!

PR #682, opened 2024-05-16
chore: hdfs - remove unnecessary jars and bump dependency versions to fix CVEs

adwk67:
# Description

This PR fixes the following CVEs for snappy-java (bumped from 1.1.8.2 to 1.1.10.4):

CVE-2023-34453
CVE-2023-34454
CVE-2023-43642
CVE-2023-34455

It also strips the image of jars that we do not need (yarn, mapreduce, minicluster modules, plus test/source jars).

```[tasklist]
```

<details>

```shell
```

</details>

### DIFF
--- a/hadoop/Dockerfile
+++ b/hadoop/Dockerfile
@@ -53,7 +53,9 @@ WORKDIR /stackable
 
 # Hadoop Pipes requires libtirpc to build, whose headers are not packaged in RedHat UBI, so skip building this module
 # Build from source to enable FUSE module, and to apply custom patches.
-# Also skip building the yarn, mapreduce and minicluster modules.
+# Also skip building the yarn, mapreduce and minicluster modules: this will result in the modules being excluded but not all
+# jar files will be stripped if they are needed elsewhere e.g. share/hadoop/yarn will not be part of the build, but yarn jars
+# will still exist in share/hadoop/tools as they would be needed by the resource estimator tool. Such jars are removed in a later step.
 RUN curl --fail -L "https://repo.stackable.tech/repository/packages/hadoop/hadoop-${PRODUCT}-src.tar.gz" | tar -xzC . && \
     patches/apply_patches.sh ${PRODUCT} && \
     cd hadoop-${PRODUCT}-src && \
@@ -155,7 +157,7 @@ ENV HADOOP_YARN_HOME=/stackable/hadoop
 ENV HADOOP_MAPRED_HOME=/stackable/hadoop
 
 # Remove unneeded binaries:
-#  - source code
+#  - code sources
 #  - mapreduce/yarn binaries that were built as cross-project dependencies
 #  - minicluster (only used for testing) and test .jars
 #  - json-io: this is a transitive dependency pulled in by cedarsoft/java-utils/json-io and is excluded in 3.4.0. See CVE-2023-34610.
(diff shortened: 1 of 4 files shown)


hadoop/Dockerfile
```diff
@@ -153,5 +154,20 @@ ENV ASYNC_PROFILER_HOME=/stackable/async-profiler
...
+#  - json-io: this is a transitive dependency pulled in by cedarsoft/java-utils/json-io and is excluded in 3.4.0. See CVE-2023-34610.
+RUN rm -rf /stackable/hadoop/share/hadoop/common/sources/ && \
+  rm -rf /stackable/hadoop/share/hadoop/hdfs/sources/ && \
+  rm -rf /stackable/hadoop/share/hadoop/tools/sources/ && \
+  rm -rf /stackable/hadoop/share/hadoop/tools/lib/json-io-*.jar && \
+  rm -rf /stackable/hadoop/share/hadoop/tools/lib/hadoop-mapreduce-client-*.jar && \
+  rm -rf /stackable/hadoop/share/hadoop/tools/lib/hadoop-yarn-server*.jar && \
+  find . -name hadoop-minicluster-*.jar -type f -delete && \
+  find . -name hadoop-client-minicluster-*.jar -type f -delete && \
+  find . -name hadoop-*tests.jar -type f -delete
```
lfrancke, line 170: There is the shellcheck warning which actually might make sense here to look at.

hadoop/Dockerfile
```diff
@@ -153,5 +154,20 @@ ENV ASYNC_PROFILER_HOME=/stackable/async-profiler
...
+#  - source code
+#  - mapreduce/yarn binaries that were built as cross-project dependencies
+#  - minicluster (only used for testing) and test .jars
+#  - json-io: this is a transitive dependency pulled in by cedarsoft/java-utils/json-io and is excluded in 3.4.0. See CVE-2023-34610.
+RUN rm -rf /stackable/hadoop/share/hadoop/common/sources/ && \
+  rm -rf /stackable/hadoop/share/hadoop/hdfs/sources/ && \
+  rm -rf /stackable/hadoop/share/hadoop/tools/sources/ && \
+  rm -rf /stackable/hadoop/share/hadoop/tools/lib/json-io-*.jar && \
+  rm -rf /stackable/hadoop/share/hadoop/tools/lib/hadoop-mapreduce-client-*.jar && \
+  rm -rf /stackable/hadoop/share/hadoop/tools/lib/hadoop-yarn-server*.jar && \
```
nightkr, line 167: Shouldn't we exclude these from building instead?

adwk67: We exclude the modules from the final build, but all the jars are built regardless of the `mvn .... -pl` directive, due to cross-dependencies. Hence the second step to strip them here.

fhennig: it looks like `-pl '!hadoop-tools/hadoop-pipes,!hadoop-yarn-project,!hadoop-mapreduce-project,!hadoop-minicluster'` should exclude at least some of these, maybe it's not working? maybe they cannot be excluded? A comment would be good in that case, explaining why they cannot be excluded from the build.

adwk67: -pl results in there being e.g. no share/hadoop/yarn or share/hadoop/mapreduce folders in the image, although various yarn/MR binaries are still needed in e.g. share/tools. I'll add a comment.

hadoop/Dockerfile
adwk67: See 878cc24.

lfrancke (dismissed):

hadoop/Dockerfile
adwk67: Comment is hopefully a bit clearer here: 878cc24

lfrancke (approved):

hadoop/Dockerfile
fhennig: Works for me

PR #694, opened 2024-05-23
Patch NiFi to allow disabling the host header check

soenkeliebau:
# Description

NiFi has the configuration option 'nifi.web.proxy.host' which controls allowed
values for the host header field in any incoming request for the web ui.

This frequently causes issues when trying to expose the NiFi UI via for example
an ingress, loadbalancer or any similar type of mechanism.

NiFi does not allow to disable this behavior, so at the moment the nifi operator
simply hardcodes all even remotely possible values into this field.
But in order to allow putting for example in ingress in front of NiFi this means
using config overrides to change the value of this option, copy all the values
the operator put in there and add the extra value you need.

This is less than ideal, the proper solution would probably be
https://github.com/stackabletech/nifi-operator/issues/604

But until that is merged this is a simple workaround that allows overriding the list of allowed
hostnames by just setting it to "*" and this will effectively bypass the hostname check entirely if set.

This allows us to keep the default behavior in place for those users where it works and not remove
security features, but also enables users to disable this check if they know what they are doing.

```[tasklist]
```

<details>

```shell
```

</details>


lfrancke:
Haven't looked at the details but please rename the patch files to contain the target version as well (see e.g. HBase and others).
That makes it easier to identify the file when it is copied around in isolation

soenkeliebau:
> Haven't looked at the details but please rename the patch files to contain the target version as well (see e.g. HBase and others). That makes it easier to identify the file when it is copied around in isolation

Is it okay to have a range of versions in there? The same patch applies across a lot of NiFi versions, that file wasn't touched a lot it seems.

nifi/stackable/patches/1.21.0/002-NIFI-no-host-header-check-1.21.0.patch
```diff
@@ -0,0 +1,58 @@
+Subject: [PATCH] Allow bypassing check for host header.
```
sbernauer, line 1: Attention: We don't build NiFi 1.21 from source, so this patch files have no effect (thx @maltesander for pointing this out).
Can you please place a README or similar in this folder instead that explains that you can not patch 1.21?

soenkeliebau: Do we plan on building it from source at some point? Otherwise I'll just remove the patch instead, no need to confuse anybody if we never plan on supporting it.

sbernauer: > Otherwise I'll just remove the patch instead

That was exactly my suggestion ;) With one addition of instead having a README stating this folder is not used

sbernauer (approved):
Thanks!

lfrancke:
Do I understand correctly that this is in preparation for adding the necessary support in the operator so there is nothing to document here?

sbernauer:
Exactly :+1:

PR #695, opened 2024-05-27
feat: Add log processing script to opa for decision logging

xeniape:
# Description

Part of https://github.com/stackabletech/opa-operator/issues/422

This PR adds the script which is used in the opa start command set by the opa-operator to filter incoming logs by log level and route logs to the correct output (file or console).

```[tasklist]
- [x] Integration tests ran successfully https://ci.stackable.tech/view/02%20Operator%20Tests%20(custom)/job/opa-operator-it-custom/96/
```

<details>

```shell
```

</details>


opa/stackable/bin/process-logs
```diff
@@ -0,0 +1,108 @@
+#!/usr/bin/env bash
```
NickLarsenNZ, line 1: I think we should run scripts in strict mode, eg:

```suggestion
#!/usr/bin/env bash
set -euo pipefail
```

opa/stackable/bin/process-logs
```diff
@@ -0,0 +1,139 @@
...
+    esac
+}
+
+main() {
+    parse_args $@
+
+    local DECISION_LEVELS=$(get_levels $DECISION_LEVEL)
+    local SERVER_LEVELS=$(get_levels $SERVER_LEVEL)
+    local CONSOLE_LEVELS=$(get_levels $CONSOLE_LEVEL)
+    local FILE_LEVELS=$(get_levels $FILE_LEVEL)
```
NickLarsenNZ, line 128: I think for safety, we should wrap the variables in quotes to disallow execution of commands.

```suggestion
    local DECISION_LEVELS=$(get_levels "$DECISION_LEVEL")
    local SERVER_LEVELS=$(get_levels "$SERVER_LEVEL")
    local CONSOLE_LEVELS=$(get_levels "$CONSOLE_LEVEL")
    local FILE_LEVELS=$(get_levels "$FILE_LEVEL")
```

There are more cases below too.

NickLarsenNZ (commented):
Nice script. If you haven't already, it might pay to run it through the shellcheck linter.

opa/stackable/bin/process-logs
```diff
@@ -0,0 +1,139 @@
...
+
+    jq -c --unbuffered 'if .decision_id then .logger = "decision" else .logger = "server" end' |
```
sbernauer, line 130: What a beauty :)

sbernauer (dismissed):
Very nice, exactly what I was thinking of! Haven't properly reviewed the operator part yet, feel free to wait with merge until both are ready to be merged.
Also happy about @siegfriedweber's review

opa/stackable/bin/process-logs
```diff
@@ -0,0 +1,139 @@
...
+        --container-name)
+            shift
+            CONTAINER_NAME=$1
+            ;;
+        *)
+            break
+            ;;
+    esac
+    shift
+  done
```
nightkr, line 60: Wouldn't it be simpler to just pass these as envvars directly?

opa/stackable/bin/process-logs
```diff
@@ -0,0 +1,139 @@
...
+main() {
+    parse_args $@
+
+    local DECISION_LEVELS=$(get_levels $DECISION_LEVEL)
+    local SERVER_LEVELS=$(get_levels $SERVER_LEVEL)
+    local CONSOLE_LEVELS=$(get_levels $CONSOLE_LEVEL)
+    local FILE_LEVELS=$(get_levels $FILE_LEVEL)
+
+    jq -c --unbuffered 'if .decision_id then .logger = "decision" else .logger = "server" end' |
+    jq -c --unbuffered --arg decision_levels $DECISION_LEVELS --arg server_levels $SERVER_LEVELS \
```
nightkr, line 131: Any reason to use separate jqs for these, rather than just piping inside of jq?

nightkr: Actually, thinking about it a bit more, this feels like exactly the sort of remapping we'd use Vector VRL for.

opa/stackable/bin/process-logs
```diff
@@ -0,0 +1,145 @@
...
+    local FILE_LEVELS
+    FILE_LEVELS=$(get_levels "$FILE_LEVEL")
+
+    jq -c --unbuffered 'if .decision_id then .logger = "decision" else .logger = "server" end' |
+    jq -c --unbuffered --arg decision_levels "$DECISION_LEVELS" --arg server_levels "$SERVER_LEVELS" \
+        'select(((.logger == "decision") and (.level | inside($decision_levels))) or 
+        ((.logger == "server") and (.level | inside($server_levels))))' |
+    tee >(jq -c --unbuffered --arg file_levels "$FILE_LEVELS" 'select(.level | inside($file_levels))' \
+        > >(/stackable/multilog s"$OPA_ROLLING_LOG_FILE_SIZE_BYTES" n"$OPA_ROLLING_LOG_FILES" "$STACKABLE_LOG_DIR"/"$CONTAINER_NAME")) |
+    jq -c --unbuffered --arg console_levels "$CONSOLE_LEVELS" 'select(.level | inside($console_levels))'
```
siegfriedweber, line 142: The pipe terminates if an error occurs. This is the case e.g. if the input is not valid JSON or the JSON object does not contain the level property. It is important that this script is fail-safe, otherwise the command calling it, will probably also fail.

This comment contradicts the proposal above that `set -euo pipefail` should be set.

siegfriedweber (changes requested):

opa/stackable/bin/process-logs
xeniape: I added something that maybe combines the two things, please take a look at it and maybe @NickLarsenNZ too.

So now `-eo pipefail` is still in there for any failures other than the processing of log messages. And jq processing is wrapped in a try-catch block to catch log messages which are missing the level property and also are invalid json (by using -R and fromjson parsing). Would that solve both concerns?

Also for some reason after applying shellcheck linting suggestions to change `main $@` and `parse_args $@` to `main "$@"` and `parse_args "$@"` the -u flag throws an unbound variable error
`./opa/stackable/bin/process-logs: line 24: $1: unbound variable` which shouldn't be true? When removing the flag and running the same command again, it works fine. Maybe someone has an idea, I removed the flag for now since the unset variables are handled anyway as mentioned earlier.

NickLarsenNZ: Just in addition, you can enable/disable the features around certain blocks of code.

```sh
set -euo pipefail

# code here

set +e
# failures ignored here
set -e

# failures not ignored anymore
```

siegfriedweber: Works perfectly. Thank you!

opa/stackable/bin/process-logs
```diff
@@ -0,0 +1,144 @@
...
+        warn)
+            echo '["warn","error","fatal"]' ;;
+        info)
+            echo '["info","warn","error","fatal"]' ;;
+        debug)
+            echo '["debug","info","warn","error","fatal"]' ;;
+        trace)
+            echo '["trace","debug","info","warn","error","fatal"]' ;;
+        *)
+            echo '[""]' ;;
```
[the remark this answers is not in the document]
NickLarsenNZ: Done in 554675f718ba2d7c5f526e83f3fed75b72e4f839

opa/stackable/bin/process-logs
NickLarsenNZ: Done in 92ed079

opa/stackable/bin/process-logs
NickLarsenNZ: `-eo` Done in b5533dbf5b94a8ae8b52bdc43a36a2bc44fcf3bf

In this case, the script is used in a redirect, eg: `opa run ... &> >(process-logs)`. So any failure will just stop the log processing and all output from `opa` will be lost (but the process will continue running).

opa/stackable/bin/process-logs
NickLarsenNZ: True, it would be simpler. Given that it is done like this already, is it a show-stopper?

opa/stackable/bin/process-logs
nightkr: Given that it's all going into the file _after_ processing and that it's going into Vector afterwards anyway, yes.

On the other hand, we'd probably want to address https://github.com/stackabletech/operator-rs/issues/794 first, so fine to punt on it for now.

PR #723, opened 2024-06-19
Add a script to upload a new protoc version

lfrancke:
# Description

Add a script to upload a new protoc version
This also changes all scripts to include "set -euo pipefail" in a consistent way


razvan (approved):
i see protoc 27.1 sources available in Nexus so I assume the script worked.

PR #725, opened 2024-06-19
Change the dev workflow so that manifest lists are built and pushed immediately

lfrancke:
# Description

Change the dev workflow so that manifest lists are built and pushed right after the images are built
If needed this can be split out even further to build every product version separately

### DIFF
--- a/.github/workflows/dev.yaml
+++ b/.github/workflows/dev.yaml
@@ -14,7 +14,6 @@ jobs:
     runs-on: ${{ matrix.runner }}
     strategy:
       # There is value to see which builds are working and which are actually broken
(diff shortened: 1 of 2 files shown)


lfrancke:
I think there might actually be a problem here and I don't _think_ it'll work because of our sharding.
If I understand it correctly we might get 5 shards where 4 do not actually build anything but they'll jump straight to the manifest part and that'll probably fail or multiple shards will run the same command....

razvan:
> multiple shards will run the same command....

lfrancke:
I'll close and delete this PR. It won't work as-is and we will work on a different solution in https://github.com/stackabletech/issues/issues/595

PR #727, opened 2024-06-20
Move the Log4j ENV variable to the java-base image

lfrancke:
# Description

ENV variables are inherited by images if they come from a parent image. This ENV variable disables the Log4j feature that caused CVE-2021-4428 (Log4Shell). Even though this probably has been disabled everywhere already it doesn't hurt to set this variable

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -47,6 +47,7 @@ All notable changes to this project will be documented in this file.
 - opa: Build from source ([#676])
 - trino: Build from source ([#687]).
 - spark: Build from source ([#679])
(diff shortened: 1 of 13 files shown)


maltesander (dismissed):
Thanks, LGTM. Could use a changelog entry though?

lfrancke:
Will do

lfrancke:
[`95d3cc5` (#727)](https://github.com/stackabletech/docker-images/pull/727/commits/95d3cc583b17e257b7569cd6d5412b69b93bdd8b)

PR #730, opened 2024-06-28
Cleanup

lfrancke:
# Description

- Cleans various caches after build to minimize size of intermediate images
- Only copy necessary stuff over (e.g. the "patches" directory was there as well as other things like maven directory)
- Removes Druid 27.0.0 which we will not support in SDP 24.7 anymore

```[tasklist]
```

<details>

```shell
```

</details>

### DIFF
--- a/druid/Dockerfile
+++ b/druid/Dockerfile
@@ -36,7 +36,11 @@ RUN curl --fail -L "https://repo.stackable.tech/repository/packages/druid/apache
     mvn clean install -Pdist -pl '!extensions-core/druid-ranger-security' -DskipTests -Dmaven.javadoc.skip=true && \
     tar -xzf /stackable/apache-druid-${PRODUCT}-src/distribution/target/apache-druid-${PRODUCT}-bin.tar.gz && \
     mv /stackable/apache-druid-${PRODUCT}-src/apache-druid-${PRODUCT} /stackable/apache-druid-${PRODUCT} && \
-    rm -rf /stackable/apache-druid-${PRODUCT}-src
+    rm -rf /stackable/apache-druid-${PRODUCT}-src && \
+    rm -rf /stackable/.m2 && \
+    rm -rf /stackable/.npm && \
+    rm -rf /stackable/.cache
+
     #  Do not remove the /stackable/apache-druid-${PRODUCT}/quickstart folder, it is needed for loading the Wikipedia
     # testdata in kuttl tests and the getting started guide.
 
@@ -91,8 +95,8 @@ RUN microdnf update && \
 USER stackable
 WORKDIR /stackable
 
-COPY --from=druid-builder /stackable /stackable
-COPY --chown=stackable:stackable druid/stackable /stackable
+COPY --from=druid-builder /stackable/apache-druid-${PRODUCT} /stackable/apache-druid-${PRODUCT}
+COPY --chown=stackable:stackable druid/stackable/bin /stackable/bin
 COPY --chown=stackable:stackable druid/licenses /licenses
 
 RUN ln -s /stackable/apache-druid-${PRODUCT} /stackable/druid && \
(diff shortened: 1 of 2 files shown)


druid/Dockerfile
```diff
@@ -36,7 +36,11 @@ RUN curl --fail -L "https://repo.stackable.tech/repository/packages/druid/apache
     mvn clean install -Pdist -pl '!extensions-core/druid-ranger-security' -DskipTests -Dmaven.javadoc.skip=true && \
     tar -xzf /stackable/apache-druid-${PRODUCT}-src/distribution/target/apache-druid-${PRODUCT}-bin.tar.gz && \
     mv /stackable/apache-druid-${PRODUCT}-src/apache-druid-${PRODUCT} /stackable/apache-druid-${PRODUCT} && \
-    rm -rf /stackable/apache-druid-${PRODUCT}-src
+    rm -rf /stackable/apache-druid-${PRODUCT}-src && \
```
Maleware, line 39: Is this necessary? 

At the copy step we only copy the binary and the bin/ folder. I think the 'src' folder shouldn't end up in the final image.

Maleware (commented):
Other then the question LGTM ( didn't build it )

druid/Dockerfile
lfrancke: We do this to make the intermediate image smaller due to the size constraints on Github Action runners

dervoeti:
LGTM as well, I built it locally and inspected the image, looks fine!

dervoeti (approved):
LGTM

druid/Dockerfile
```diff
@@ -36,7 +36,11 @@ RUN curl --fail -L "https://repo.stackable.tech/repository/packages/druid/apache
...
+    rm -rf /stackable/.m2 && \
+    rm -rf /stackable/.npm && \
+    rm -rf /stackable/.cache
```
razvan, line 42: Actually, now that we copy select subfolders from `/stackable ` in the next stage, it's counter-productive to  delete these.
Instead we should mount them as cache volumes so that we can reuse them on sequential builds like this:

```
RUN --mount=type=cache,target=/stackable/.m2 \
...
mvn clean install -Pdist -pl '!extensions-core/druid-ranger-security' -DskipTests -Dmaven.javadoc.skip=true
...
```

lfrancke: So a) I didn't know about that mount thing, thank you but b) I did all of this to shrink the intermediate images to not run into any github size limitations.

I'm not tied to this and happy to accept a followup PR

razvan: > I did all of this to shrink the intermediate images

`rm`-ing will not reduce the intermediate image size

lfrancke: I'm pretty sure it will. Otherwise the image is kept around until it can be used in the final image and at that point both are there.
So anything we can do to decrease the size is good.

razvan: > I'm pretty sure it will.

No it won't. The image size can only be decreased by starting a new stage not by deleting files.

Here is an example from the hbase-opa-authorizer: https://gist.github.com/razvan/021b4d08310c2aaec2f84db1a19bc70c

The image size stays the same regardless whether the Maven cache is removed or not:
```
rm-tmp-opa-auth    latest e1089835de99   5 seconds ago    1.04GB
tmp-opa-auth   latest 8293dc780586   21 minutes ago   1.04GB
```

lfrancke: Yes, that is because they are in two separate `RUN` statements. Each RUN starts a new layer.

Here however we have it all in the same layer. So...at the beginning we start a layer, do a bunch of things e.g. kick off a maven build which downloads a shitton of stuff and then delete the cache again and the end result of that is committed to a layer.

If we did not delete the stuff at the end of the RUN statement all the cache stuff would also be written to disk permanently.

I just tried. The intermediate image is ~6GB if the caches are not removed and ~1.4GB with everything removed.

razvan: Yes, you are right. I had this realization this morning.

PR #738, opened 2024-07-03
Pushing multi arch images to Nexus and Harbor

Maleware:
# Description

Adds multi architecture for mirror.yaml

```[tasklist]
```

<details>

```shell
```

</details>


Maleware:
Test succeeded: https://github.com/stackabletech/docker-images/actions/runs/9791307623/job/27034818090

The path for `csi-provisioner` and `csi-node-driver-registrar` within nexus changed as it was hard to automate for various dependencies. We will need a fix at those parts

Maleware:
Updating Helm charts for [secret-operator](https://github.com/stackabletech/secret-operator/pull/455) and [listener-operator](https://github.com/stackabletech/listener-operator/pull/203)

.github/workflows/mirror.yaml
```diff
@@ -9,36 +9,95 @@ on:
...
+        with:
+          registry: ${{ env.DST_REGISTRY_NEXUS }}
+          username: github
+          password: ${{ secrets.NEXUS_PASSWORD }}
+      - uses: sigstore/cosign-installer@59acb6260d9c0ba8f4a2f9d9b48431a222b68e20 # v3.5.0
+      - name: Build and push manifest list
+        run: |
+          # Create manifest list for Nexus and Harbor
+          docker manifest create "$DST_REGISTRY_HARBOR/${{ github.event.inputs.harbor_proj }}/${{ github.event.inputs.dst }}" --amend "$DST_REGISTRY_HARBOR/${{ github.event.inputs.harbor_proj }}/${{ github.event.inputs.dst }}-arm64" --amend "$DST_REGISTRY_HARBOR/${{ github.event.inputs.harbor_proj }}/${{ github.event.inputs.dst }}-amd64"
+          docker manifest create "$DST_REGISTRY_NEXUS/$DST_PROJECT_NEXUS/${{ github.event.inputs.dst }}" --amend "$DST_REGISTRY_NEXUS/$DST_PROJECT_NEXUS/${{ github.event.inputs.dst }}-arm64" --amend "$DST_REGISTRY_NEXUS/$DST_PROJECT_NEXUS/${{ github.event.inputs.dst }}-amd64"
```
dervoeti, line 97: You refer to the images by tag name here, which is not a huge problem, but a race condition might occur:
If the x86 image is built before the ARM image and the build of the ARM image takes a long time, the x86 image might get overridden in the meantime (very unlikely to happen in practice, but possible in theory). That means that the manifest list could potentially point to images with different versions for x86 and ARM.

Since we have the image digests in the build step anyway (needed for signing) you could pass them to the "Create and push manifest list" job as input and then refer to the images by digest when creating the manifest list.

I think it's not a huge problem, just wanted to point it out. If it's not too complicated to fix than I'd prefer that solution since it's the correcter one. If it's too much effort I'm fine with the current solution as well.

Maleware: @dervoeti done with https://github.com/stackabletech/docker-images/pull/738/commits/6c245df3d15b0c25b0906e7a50041f6911121cbb 

thanks

dervoeti (approved):
LGTM

Maleware:
There will be a follow up with https://github.com/stackabletech/docker-images/issues/785 as we want everything fit into our reusable workflows.

PR #733, opened 2024-07-02
feat!: Split up CI workflow per product

Techassi:
Part of https://github.com/stackabletech/issues/issues/595

This PR splits up the `dev` workflow so that each individual product image can be built instead of building all products at once. This allows us to trigger workflows on changes in specific paths. Additionally, it is now possible to split up the conf.py file, see #732 
 
| Before | After |
| -- | -- |
| image | image |

```[tasklist]
- [ ] Changes are OpenShift compatible
- [ ] All added packages (via microdnf or otherwise) have a comment on why they are added
- [ ] Things not downloaded from Red Hat repositories should be mirrored in the Stackable repository and downloaded from there
- [ ] All packages should have (if available) signatures/hashes verified
- [ ] Add an entry to the CHANGELOG.md file
- [ ] Integration tests ran successfully
```

PR #763, opened 2024-07-08
fix(ci): Re-add free-disk-space action

Techassi:
This cleans up some space on the runner to be able to build products which produce a bunch of data during build. This action was present before, but accidentally removed in #733. This now re-adds the action to fix the larger failing builds.


NickLarsenNZ (approved):
LGTM

PR #764, opened 2024-07-08
refactor: Move scripts into .scripts folder

Techassi:
This PR moves all scripts into a dedicated `.scripts` folder to clean up the repository root.


lfrancke:
I know it's still a draft, but: Should it really be a hidden folder?
I like the moving out in general but would suggest `scripts` instead.
But I'll leave the final decision up to you.

Techassi:
> I know it's still a draft, but: Should it really be a hidden folder? I like the moving out in general but would suggest `scripts` instead. But I'll leave the final decision up to you.

@NickLarsenNZ and I discussed this as well. We would like to avoid `scripts` because it will sit between product folders, which is not something we want. To "push it to the top" of the file tree, we thought about using `.` or `_` as a prefix. Since other well-known folders already use `.` we did the same.

lfrancke:
Thanks!

NickLarsenNZ:
I don't want to muddy the waters further, but since @lfranke mentioned it, my preference would be to move the products to a folder, and then I don't mind if the scripts move to a normal `scripts/` folder.

My reasoning is more about presentation in Github, it takes way too long to scroll down to see the beautify README :sweat_smile:

lfrancke:
I don't have an opinion on that suggestion. Neither positive nor negative. Decide on your own :)

Techassi:
> I don't want to muddy the waters further, but since @lfrancke mentioned it, my preference would be to move the products to a folder, and then I don't mind if the scripts move to a normal `scripts/` folder.
> 
> My reasoning is more about presentation in Github, it takes way too long to scroll down to see the beautify README 😅

Yes, that is definitely a good alternative. But that would require a bigger change which is out of scope for this PR in my opinion.

NickLarsenNZ:
> > I don't want to muddy the waters further, but since @lfrancke mentioned it, my preference would be to move the products to a folder, and then I don't mind if the scripts move to a normal `scripts/` folder.
> > My reasoning is more about presentation in Github, it takes way too long to scroll down to see the beautify README 😅
> 
> Yes, that is definitely a good alternative. But that would require a bigger change which is out of scope for this PR in my opinion.

I would say its the alternative to this PR. I would like to eventually do it, but this is fine to go in too.

NickLarsenNZ (approved):

PR #768, opened 2024-07-09
Add versions 3.1.3 and 4.0.2

maltesander:
# Description

- added: 3.1.3, 4.0.1
- removed: 2.1.1 , 3.0.1, 3.0.3

Query export fix patch already in 3.1.3 and 4.0.2 :)

```[tasklist]
```

<details>

```shell
```

</details>


superset/versions.py
```diff
@@ -14,24 +7,24 @@
...
         "python": "3.9",
         "vector": "0.39.0",
         "statsd_exporter": "0.26.1",
-        "authlib": "1.2.1",  # https://github.com/dpgaspar/Flask-AppBuilder/blob/v4.3.10/requirements-extra.txt#L7
+        "authlib": "1.2.1",  # https://github.com/dpgaspar/Flask-AppBuilder/blob/release/4.4.1/requirements/extra.txt#L7
     },
     {
-        "product": "3.1.0",
+        "product": "4.0.2",
         "python": "3.9",
```
Techassi, line 25: question: Do we want to use Python 3.10 or even 3.11 here? It seems like Superset supports both newer versions.

maltesander: I think 3.9 to 3.11 is officially supported

maltesander: i can try 3.11 locally

maltesander: 3.10 and 3.11 do not work out of the box. Would need adaptations of the docker file.

Techassi: We decided to tackle the Python bump at a later point in time. Merging as is.

Techassi (approved):

PR #777, opened 2024-07-12
Build statsd_exporter from source

### LINKED ISSUE #776: Build statsd_exporter from source
Building `statsd_exporter` from source enables manual dependency bumps and the fact the we compile `statsd_exporter` with a newer Go version alone should already reduce a lot of vulnerabilities in `statsd_exporter`, since the `stdlib` is updated.

dervoeti:
# Description

Fixes #776

This PR enables building `statsd_exporter`, which is used in our Airflow and Superset images, from source.

I moved the build process of `statsd_exporter` into its own image, similar to Vector, to reduce code duplication and improve caching (e.g. if it's built for Airflow the built `statsd_exporter` binary in the cache can be reused in the Superset build).

I spun up a local cluster with the new Airflow image and inspected the `metrics` container, it provides metrics on port 9102. I also scanned the new image and it contains less vulnerabilities (this removed 2 of 3 critical vulnerabilities plus a few others).

I built the Airflow 2.9.2 image with the changes and ran the integration tests locally, they all passed.


statsd_exporter/Dockerfile
```diff
@@ -0,0 +1,29 @@
...
+        gzip \
+        # used to build statsd_exporter
+        golang && \
+    microdnf clean all && \
+    rm -rf /var/cache/yum && \
+    mkdir -p /statsd_exporter
+WORKDIR /statsd_exporter
+
+# hadolint ignore=DL3003
+RUN curl --fail -L "https://repo.stackable.tech/repository/packages/statsd_exporter/statsd_exporter-${PRODUCT}.src.tar.gz" | \
```
razvan, line 27: The current best practice is to use heredoc scripts and mount the dependency volumes as caches. An example is here: https://github.com/stackabletech/docker-images/blob/71868dae6d698974e50033c13049f452a094992a/hbase/Dockerfile#L44

razvan (commented):
A changelog entry is missing plus a comment.

statsd_exporter/Dockerfile
dervoeti: Thanks! I addressed the issues.
I also spawn a subshell now instead of `cd`-ing in and out of the source directory because hadolint complained about this ([SC2103](https://github.com/koalaman/shellcheck/wiki/SC2103)).

statsd_exporter/Dockerfile
```diff
@@ -0,0 +1,35 @@
...
+WORKDIR /statsd_exporter
+
+# hadolint ignore=DL3003
+RUN --mount=type=cache,id=go-statsd-exporter,uid=1000,target=/go_cache <<EOF
+export GOPATH=/go_cache
+curl --fail -L "https://repo.stackable.tech/repository/packages/statsd_exporter/statsd_exporter-${PRODUCT}.src.tar.gz" | tar -xzC .
+(
+    cd "statsd_exporter-${PRODUCT}" || exit
+    go build -o ../statsd_exporter
+)
```
razvan, line 33: Not sure why the `exit` and the parens were needed...

```suggestion
    cd "statsd_exporter-${PRODUCT}"
    go build -o ../statsd_exporter
```

razvan (approved):
lgtm! (made a suggestion though)

statsd_exporter/Dockerfile
dervoeti: That's the subshell, so the `cd` only affects the commands in the subshell (only `go build` in that case). The commands after the parentheses are executed in the directory we were in before (which is needed for the `rm` to work as intended). As mentioned, I only introduced this because hadolint complained. Before I just did `cd ..` after the `go build`.

dervoeti:
I also changed the package installation step to use heredoc

statsd_exporter/Dockerfile
razvan: Ah, got it! So the `ignore=DL3003` comment can be removed?

razvan:
> I also changed the package installation step to use heredoc

Don't want to  be a nag... at this point you could also put everything in one RUN layer.

dervoeti:
> > I also changed the package installation step to use heredoc
> 
> Don't want to be a nag... at this point you could also put everything in one RUN layer.

Oh right! Done.

PR #778, opened 2024-07-16
fix(airflow): path to libstdc++.so

razvan:
# Description

When starting a container from the Airflow image, the following error is shown:

```shell
$ docker run -ti -u 1000:0 --rm docker.stackable.tech/sandbox/airflow:2.9.2-stackable0.0.0-dev bash
ERROR: ld.so: object '/usr/lib/x86_64-linux-gnu/libstdc++.so.6' from LD_PRELOAD cannot be preloaded (cannot open shared object file): ignored.
...
```

The problem is that on UBI that path wrong and that the g++ package needs to be installed.

```[tasklist]
```

<details>

```shell
```

</details>

### DIFF
--- a/airflow/Dockerfile
+++ b/airflow/Dockerfile
@@ -77,13 +77,11 @@ LABEL name="Apache Airflow" \
 COPY airflow/licenses /licenses
 
 # Update image and install python
-# g++ is required for libstdc++ that is preloaded in entrypoint.sh
 RUN microdnf update && \
     microdnf install \
     ca-certificates \
     cyrus-sasl \
     git \
-    g++ \
     libpq \
     openldap \
     openldap-clients \
--- a/airflow/Dockerfile
+++ b/airflow/Dockerfile
@@ -77,11 +77,13 @@ LABEL name="Apache Airflow" \
 COPY airflow/licenses /licenses
 
 # Update image and install python
+# g++ is required for libstdc++ that is preloaded in entrypoint.sh
 RUN microdnf update && \
-    microdnf install  \
+    microdnf install -y \
     ca-certificates \
     cyrus-sasl \
     git \
+    g++ \
     libpq \
     openldap \
     openldap-clients \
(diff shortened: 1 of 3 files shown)


maltesander (approved):
LGTM!

airflow/Dockerfile
```diff
@@ -77,11 +77,13 @@ LABEL name="Apache Airflow" \
 COPY airflow/licenses /licenses
 
 # Update image and install python
+# g++ is required for libstdc++ that is preloaded in entrypoint.sh
 RUN microdnf update && \
-    microdnf install  \
+    microdnf install -y \
```
dervoeti, line 82: I think we don't need the `-y` here since we have `assumeyes=True` in `dnf.conf`

razvan: deleted again https://github.com/stackabletech/docker-images/pull/778/commits/d6c7fc55b1e94b519d4f3d0d87939525228a826d

razvan:
> @razvan Are you sure we need the `g++` package? I just ran the 2.9.2 dev image and `/usr/lib64/libstdc++.so.6` seems to be already present.

Hm, you are right. I started from /usr/lib/... and there was nothing there untill g++ was installed.

Reverted the g++ installation

airflow/stackable/utils/entrypoint.sh
```diff
@@ -34,7 +34,9 @@ set -euo pipefail
 # The side effect of this is slightly (in the range of 100s of milliseconds) slower load for any
 # binary started and a little memory used for Heap allocated by initialization of libstdc++
 # This overhead is not happening for binaries that already link dynamically libstdc++
-LD_PRELOAD="/usr/lib/$(uname -m)-linux-gnu/libstdc++.so.6"
+# LD_PRELOAD="/usr/lib/$(uname -m)-linux-gnu/libstdc++.so.6"
+# Stackable: this requires the g++ dnf package to be installed in the image.
```
dervoeti, line 38: That comment can be removed as well and I think then we're good :slightly_smiling_face: Maybe remove the commented line above this one as well.

razvan: Done. I left the old path for future reference.

dervoeti (approved):
LGTM

PR #780, opened 2024-07-18
Move base image to new heredoc format

lfrancke:
# Description

Move base image to new heredoc format
There are no functional changes

### DIFF
--- a/stackable-base/Dockerfile
+++ b/stackable-base/Dockerfile
@@ -9,21 +9,25 @@ ENV RUST_DEFAULT_TOOLCHAIN_VERSION=1.79.0
 ENV CARGO_CYCLONEDX_CRATE_VERSION=0.4.0
 ENV CARGO_AUDITABLE_CRATE_VERSION=0.6.4
 
-RUN microdnf update --assumeyes && \
-    microdnf --assumeyes install \
-    # Needed to fetch source
-    git \
-    # Needed for compilation
-    gcc && \
-    microdnf clean all
+RUN <<EOF
+microdnf update --assumeyes
 
-RUN curl --proto '=https' --tlsv1.2 -sSf https://sh.rustup.rs | sh -s -- -y --default-toolchain $RUST_DEFAULT_TOOLCHAIN_VERSION && \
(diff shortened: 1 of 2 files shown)


stackable-base/Dockerfile
```diff
@@ -49,63 +53,86 @@ SHELL ["/bin/bash", "-euo", "pipefail", "-c"]
...
-    echo "alias ..='cd ..'" >> /stackable/.bashrc && \
-    echo "export PS1='\u@\[\e[36m\]\H\[\e[m\] \[\e[32m\]\$(pwd)\[\e[m\] \\$ '" >> /stackable/.bashrc && \
-    echo -e "if [ -f ~/.bashrc ]; then\n\tsource ~/.bashrc\nfi" >> /stackable/.profile && \
-    chown stackable:stackable /stackable/.bashrc && \
-    chown stackable:stackable /stackable/.profile
+
+# echo won't expand escape sequences. Consider printf.
+# https://github.com/koalaman/shellcheck/wiki/SC2028
+# Reason: This is complaining about the complicated PS1 statement.
+# While it seems like black it seems to work as intended so I'm not going to touch it!
```
razvan, line 60:
```suggestion
# It seems to work as intended so I'm not going to touch it!
```

stackable-base/Dockerfile
```diff
@@ -49,63 +53,86 @@ SHELL ["/bin/bash", "-euo", "pipefail", "-c"]
...
+  nano \
+  tar
+
+# Added only temporarily to create the user and group, removed again below
+microdnf install shadow-utils
+groupadd --gid 1000 --system stackable
+useradd --gid stackable --uid 1000 --system stackable -d /stackable
+
+mkdir /stackable
+chown stackable:stackable /stackable
```
razvan, line 99: I always wondered why we don't properly assign `/stackable` as home dir to the `useradd` command and let it do the right thing.

lfrancke: I can try to look at that in a follow-up PR. I wanted this to be purely "formatting" changes.

razvan (approved):

stackable-base/Dockerfile
lfrancke: It looks like we already do and these two lines might actually be redundant. I'll check!

PR #782, opened 2024-07-18
Adds more global hadolint ignores with reasoning

lfrancke:
# Description

Adds more global hadolint ignores with reasoning and moves most hadolint ignores to a global config file so we have to define them only once.


NickLarsenNZ (approved):
LGTM

PR #789, opened 2024-07-23
Fix release workflow

lfrancke:
# Description

- The enumerate-product-versions.py would not work because it can't find the module "conf" since it was moved to the .scripts subfolder

- Changes the release workflow to fail when the enumerate script fails. It used to just swallow the error

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -191,7 +191,7 @@ jobs:
             echo "Python script failed"
             exit 1
           fi
-          
+
           for product_and_version in $product_and_versions; do
             PRODUCT="$(echo "$product_and_version" | cut -d '#' -f 1)"
             VERSION="$(echo "$product_and_version" | cut -d '#' -f 2)"
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -97,10 +97,10 @@ jobs:
           # Bake images and load them to local docker repo for signing
           # TODO: buildx cannot --load and --push at the same time
           # Tagging images with the architecture they were build on
-          ARCH_FOR_DOCKER="$(arch | sed -e 's#x86_64#amd64#' | sed -e 's#aarch64#arm64#')"
+          ARCH="$(uname -m | sed -e 's#x86_64#amd64#' | sed -e 's#aarch64#arm64#')"
           bake --product "${{ matrix.product }}" \
-          --image-version "${GITHUB_REF_NAME}-${ARCH_FOR_DOCKER}" \
-          --architecture "linux/${ARCH_FOR_DOCKER}" \
+          --image-version "${GITHUB_REF_NAME}-${ARCH}" \
+          --architecture "linux/${ARCH}" \
           --organization stackable \
           --shard-count "${{matrix.shard_count}}" \
           --shard-index "${{matrix.shard_index}}" \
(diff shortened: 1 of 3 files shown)


.github/workflows/release.yml
```diff
@@ -97,10 +97,10 @@ jobs:
           # Bake images and load them to local docker repo for signing
           # TODO: buildx cannot --load and --push at the same time
           # Tagging images with the architecture they were build on
-          ARCH_FOR_DOCKER="$(arch | sed -e 's#x86_64#amd64#' | sed -e 's#aarch64#arm64#')"
+          ARCH="$(uname -m | sed -e 's#x86_64#amd64#' | sed -e 's#aarch64#arm64#')"
           bake --product "${{ matrix.product }}" \
-          --image-version "${GITHUB_REF_NAME}-${ARCH_FOR_DOCKER}" \
-          --architecture "linux/${ARCH_FOR_DOCKER}" \
+          --image-version "${GITHUB_REF_NAME}-${ARCH}" \
+          --architecture "linux/${ARCH}" \
```
NickLarsenNZ, line 103: Thanks you :smile:

.scripts/enumerate-product-versions.py
```diff
@@ -1,3 +1,9 @@
+import sys
+import os
+
+# Add parent directory to the sys.path
+sys.path.append(os.path.dirname(os.path.dirname(os.path.abspath(__file__))))
```
NickLarsenNZ, line 5: Does it need to be the parent dir or should it be the CWD?

We did the CWD version here: https://github.com/stackabletech/image-tools/pull/27/files

```suggestion
# Add CWD directory to the sys.path
sys.path.append(str(os.getcwd()))
```

NickLarsenNZ (approved):

.scripts/enumerate-product-versions.py
lfrancke: "my" version works regardless of whether you call the script from / or from ./scripts.
the cwd version only works if being called from /

but the cwd version will continue working as long as the script is always called from the folder where `conf.py` resides, my version will continue working as long as the relative location of the two files stays the same (parent/child)

I'd say both are not perfect and there's no clear winner.

NickLarsenNZ: If it works :shrug:

NickLarsenNZ (approved):
Still don't like the sys.path thing not being CWD, but that can be fixed later.

.scripts/enumerate-product-versions.py
NickLarsenNZ: @lfrancke, I checked locally, and indeed the cwd way works.

image

```diff
diff --git a/.scripts/enumerate-product-versions.py b/.scripts/enumerate-product-versions.py
index d4afe83..08c29e7 100644
--- a/.scripts/enumerate-product-versions.py
+++ b/.scripts/enumerate-product-versions.py
@@ -1,3 +1,8 @@
+import os
+import sys
+
+sys.path.append(str(os.getcwd()))
+
 import conf
 
 PRODUCTS = [
```

lfrancke: Yes. I know. But it only works if called from `/` (the root directory of the repo)

The other solutions also works when you e.g. call it from within the `.scripts` directory while the `cwd` version doesn't.

PR #790, opened 2024-07-24
ci(release): Various workflow fixes

NickLarsenNZ:
- [x] Get the container digest from structured output
    - [x] New publish action
    - [x] Release workflow
- [x] Make cargo quiet
- [x] fix pre-commit to compare against main (https://github.com/stackabletech/issues/issues/616)

### DIFF
--- a/.github/actions/publish-image/action.yml
+++ b/.github/actions/publish-image/action.yml
@@ -63,14 +63,9 @@ runs:
       shell: bash
       run: |
         set -euo pipefail
-        # Store the output of `docker image push` into a variable, so we can
-        # parse it for the digest
-        PUSH_OUTPUT=$(docker image push "$(< bake-target-tags)" 2>&1)
-        echo "$PUSH_OUTPUT"
-        # Obtain the digest of the pushed image from the output of `docker image
-        # push`, because signing by tag is deprecated and will be removed from
-        # cosign in the future
-        DIGEST=$(echo "$PUSH_OUTPUT" | awk "/: digest: sha256:[a-f0-9]{64} size: [0-9]+$/ { print \$3 }")
+        docker image push "$(< bake-target-tags)"
+        # Obtain the digest of the image, because signing by tag is deprecated and will be removed from cosign in the future
+        DIGEST=$(docker images --digests "$(< bake-target-tags)" --format '{{.Digest}}')
         echo "DIGEST=$DIGEST" >> $GITHUB_ENV
         # Refer to image via its digest (docker.stackable.tech/stackable/airflow@sha256:0a1b2c...)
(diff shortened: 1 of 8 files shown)


maltesander (approved):
LGTM!

sbernauer:
@dervoeti will review as well

dervoeti:
When I build an image locally with bake, push it and then run `docker images --digests "$(< bake-target-tags)" --format '{{.Digest}}'`, it prints `<none>`. But that might be because my local build was in some way different from what we do in CI.
Also, I think `.Digest` is the local digest and the digest we are interested in is actually in `.RepoDigests`, which is an array (it should only have one element in our case though). As far as I remember, the "remote" digest in the repo you pushed to is different from the local one.

NickLarsenNZ:
I would have to look into the Digest meaning. When I ran it, it was the same as the remote, and I never saw arrays of digests, only the final.

Maybe for now, I can keep it the current way, and just print the output of this command (or even compare it), and we can check it over time to see if it matched.

dervoeti:
Sounds good! Just want to make sure that we use the correct digest, otherwise signings / SBOMs will break for new builds. If it's the same digest then it's fine for me, printing the digests to compare them sounds like a good next step. I just remembered that I saw some differences when I tried it that way, but that was about a year ago, things might have changed.

NickLarsenNZ:
On second thought @dervoeti, I'd like to keep my change. I took another look, and it appears to be fine.

The digest used to come from the last line of this output:

```
Using default tag: latest
The push refers to repository [docker.stackable.tech/sandbox/nick/alpine]
78561cef0761: Preparing
78561cef0761: Pushed
latest: digest: sha256:eddacbc7e24bf8799a4ed3cdcfa50d4b88a323695ad80f317b6629883b2c2a78 size: 528
```

This matches what I get from:

```
❯ docker images --digests docker.stackable.tech/sandbox/nick/alpine --format '{{.Digest}}'                                         
sha256:eddacbc7e24bf8799a4ed3cdcfa50d4b88a323695ad80f317b6629883b2c2a78
```

_This errors: `docker images --digests docker.stackable.tech/sandbox/nick/alpine --format '{{.RepoDigests}}'`_.

dervoeti:
Yep okay, fine with me. Maybe we can run the build once with the new workflow for one of the images? For example: https://github.com/stackabletech/docker-images/actions/workflows/dev_opa.yaml?query=branch%3Afixes-for-the-jilted-generation

NickLarsenNZ:
@dervoeti we can test it on https://github.com/stackabletech/docker-images/pull/797

dervoeti (approved):
LGTM. I'll check some signatures and SBOMs after the builds are done to verify it works as intended.

PR #797, opened 2024-08-05
chore(opa): add 0.67.1

NickLarsenNZ:
See: https://github.com/open-policy-agent/opa/releases/tag/v0.67.1

```[tasklist]
### Dependencies
- [ ] Vector product update
```

### DIFF
--- a/opa/versions.py
+++ b/opa/versions.py
@@ -6,7 +6,7 @@ versions = [
         "stackable-base": "1.0.0",
     },
     {
-        "product": "0.67.0",
+        "product": "0.67.1",
         "vector": "0.39.0",
         "bundle_builder_version": "1.1.2",
         "stackable-base": "1.0.0",
--- a/opa/versions.py
+++ b/opa/versions.py
@@ -1,11 +1,5 @@
 versions = [
     {
-        "product": "0.61.0",
-        "vector": "0.39.0",
-        "bundle_builder_version": "1.1.2",
-        "stackable-base": "1.0.0",
-    },
-    {
         "product": "0.66.0",
         "vector": "0.39.0",
         "bundle_builder_version": "1.1.2",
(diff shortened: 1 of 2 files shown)


NickLarsenNZ:
Should we now remove 0.61.0 since it was deprecated in the last release?

lfrancke (approved):

opa/versions.py
```diff
@@ -1,12 +1,12 @@
 versions = [
     {
-        "product": "0.61.0",
+        "product": "0.66.0",
         "vector": "0.39.0",
```
NickLarsenNZ, line 4: I think we need to update Vector first (if it gets updated)

sbernauer: IMHO it's fine to first bump OPA and vector afterwards

sbernauer (approved):

opa/versions.py
NickLarsenNZ: Oh yeah, it will just be a new nightly to replace the one created by this.

_I didn't see your review, and updated the description to add Vector as a dependency. Will leave it there until I come back to this (maybe tomorrow)._

sbernauer: Honestly, I didn't fully get it. I leave it to you, I'm happy to either merge or wait :)

NickLarsenNZ: I was just agreeing with you that it doesn't matter if they're done separately (this early), because:

1. Once this is merged, it will build the dev image with old Vector
2. Once Vector is updated, a new nightly will be built since the dev workflow would trigger it:
   https://github.com/stackabletech/docker-images/blob/056274f976de333caeb9ac9be4b3b8686109131f/.github/workflows/dev_opa.yaml#L18

The last line in _italics_ was just meaning I will wait because it was late and I didn't want to rush a merge. Now I'm happy to merge it and let it rebuild once Vector gets updated.

NickLarsenNZ:
I ran `./upload_new_opa_version.sh 0.67.1 ...`

PR #799, opened 2024-08-06
feat(superset): Add needed Python libs for OIDC

Maleware:
# Description

This adds a custom manager which:

1. Replaces the login button from superset with a direct link to keycloak ( login with keycloak ) and redirects to the product
2. Replaces the logout button with a direct link to keycloak which logs out and returns to superset ( Session terminated ) 
3. Adds `Flask_OIDC` and `Flask_openID` to the Docker image if it's product version `4.0.2` ( needed by fix ) 

Additionally there are going to be changes for the operator https://github.com/stackabletech/superset-operator/pull/530

This is a little more complicated then expected but works the way outlined in `001-superset_logout.patch`.

```[tasklist]
```

<details>

```shell
```

</details>


lfrancke:
Quick question: Is this needed for _everyone_?
I don't want to fix a problem for one customer just for it to break for everyone else.

Maleware:
Yes it's a generic issue mentioned here https://github.com/apache/superset/issues/23443 ( although that's Oauth ) This fixes the exact same problem for OIDC. `The CustomSecurityManager` will only be used if set in `superset_config.py` which means this is a opt-in feature and won't break anything for others ( if not willingly set )

Maleware:
Converted to draft as we want to discuss a bit more about this. To be picked up at some point.

sbernauer:
I think this waits on the decision https://github.com/stackabletech/decisions/issues/21, putting it back to progress

Maleware:
Further more, this waits on a discussion when @lfrancke is back since it's security relevance.

superset/Dockerfile
```diff
@@ -56,6 +56,11 @@ RUN python3 -m venv /stackable/app \
         pydruid \
         python-ldap \
         trino[sqlalchemy] \
+        # Enables OIDC with superset. Since https://github.com/stackabletech/superset-operator/pull/530
+        # user can inject arbitrary python code into superset_conf.py. Thus OIDC configs can be changed
+        # and we want to offer necessary libraries
+        Flask_OIDC==2.2.0 \
+        Flask-OpenID==1.3.1\
```
NickLarsenNZ, line 63: If I understand it correctly, this doesn't enable OIDC, but adds dependencies for anyone who wants to opt-in, right?

If so, the comment needs revising. Maybe something like:

```suggestion
        # Add optional dependencies for use in custom Superset configurations.
        # Since https://github.com/stackabletech/superset-operator/pull/530
        # users can add custom configuration to superset_conf.py.
        Flask_OIDC==2.2.0 \
        Flask-OpenID==1.3.1\
```

NickLarsenNZ (approved):
Left a comment with a suggestion, but marked as approved in case my interpretation is wrong.

NickLarsenNZ (dismissed):
LGTM

superset/Dockerfile
```diff
@@ -38,6 +38,9 @@ RUN microdnf update \
     && microdnf clean all && \
     rm -rf /var/cache/yum
 
+# DL3042 gets ignored, since we use --no-chache-dir, hadolint can't interfere properly
+# SC2102 informs about [abba] = [ab], but we can igonre this for [sqlalchemy] https://www.shellcheck.net/wiki/SC2102
+# hadolint ignore=DL3042,SC2102
```
NickLarsenNZ, line 43:
```suggestion
# DL3042 false-positive, --no-chache-dir is specified a few lines below.
#        See https://github.com/hadolint/hadolint/issues/1042.
# hadolint ignore=DL3042
```

NickLarsenNZ: Re: `SC2102`, We should instead quote the offending line, because if globbing is enabled, it will break.

Like:

```diff
- trino[sqlalchemy] \
+ 'trino[sqlalchemy]' \
```

NickLarsenNZ (changes requested):
Just a minor change required

NickLarsenNZ (approved):
LGTM

PR #801, opened 2024-08-07
ci(dev): Extract digests from structured output

NickLarsenNZ:
# Description

- Revert bad `release.yml` workflow from #790.
  - The digest is empty, even after push (works on my machine though).
  - I have decided not to update it in this PR, as it can be improved when it uses the reusable actions. Mostly because some variable names were used that have different meanings, and it would be an easy place for errors to creep in.
- Revert bad `publish-image` action from #790.
  - The digest is empty, even after push (works on my machine though).
- Revert debug change from #800
- Extract repo digest from structured data (for the `publish-image` action).
  - It is only available after push
- Extract repo digest from structured data (for the `publish-manifest` action).
  - It _might_ only be available after push
  - Uses `docker buildx imagetools inspect` to get the digest. I checked that it matched when doing this process locally.

<details>
<summary>🔴 → 🟢</summary>

image

</details>

```[tasklist]
```

<details>

```shell
```

</details>

### DIFF
--- a/.github/actions/publish-image/action.yml
+++ b/.github/actions/publish-image/action.yml
@@ -59,69 +59,124 @@ runs:
         echo "IMAGE_NAME=$IMAGE_NAME" >> $GITHUB_ENV
         echo "TAG_NAME=$TAG_NAME" >> $GITHUB_ENV
 
+    # TODO (@NickLarsenNZ): Make this a reusable action in .github/actions/push-and-sign/action.yml
     - name: Push Image to repo.stackable.tech and sign via cosign
       shell: bash
       run: |
         set -euo pipefail
-        # Store the output of `docker image push` into a variable, so we can
-        # parse it for the digest
-        PUSH_OUTPUT=$(docker image push "$(< bake-target-tags)" 2>&1)
-        echo "$PUSH_OUTPUT"
-        # Obtain the digest of the pushed image from the output of `docker image
-        # push`, because signing by tag is deprecated and will be removed from
-        # cosign in the future
-        DIGEST=$(echo "$PUSH_OUTPUT" | awk "/: digest: sha256:[a-f0-9]{64} size: [0-9]+$/ { print \$3 }")
-        echo "DIGEST=$DIGEST" >> $GITHUB_ENV
+        TARGET_TAG="$(< bake-target-tags)"
+        docker image push "$TARGET_TAG"
+        # Obtain the digest of the image, because signing by tag is deprecated and will be removed from cosign in the future
+        # This must be done after a push, otherwise the repo digest is empty.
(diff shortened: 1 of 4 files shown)


NickLarsenNZ:
Test run here: https://github.com/stackabletech/docker-images/actions/runs/10290906085

.github/actions/publish-manifest/action.yml
```diff
@@ -45,14 +45,25 @@ runs:
...
       run: |
         set -euo pipefail
+
+        # Get the image index digest
+        # Eg: index_digest docker.stackable.tech/stackable/hello-world:0.0.1-SNAPSHOT-stackable0.0.0-dev
+        #
+        # Note: `docker manifest push` currently outputs the same hash, but `manifest` is experimental and the
+        # STDOUT is more likely to change than the structured output.
+        function index_digest {
+          docker buildx imagetools inspect --format '{{json .Manifest}}' "$1" | jq -r '.digest'
```
dervoeti, line 55:
```suggestion
          docker buildx imagetools inspect --format '{{println .Manifest.Digest}}' "$1"
```

This can be simplified

.github/actions/publish-image/action.yml
```diff
@@ -59,62 +59,124 @@ runs:
...
+          docker inspect "$TARGET_TAG" --format json | \
+          jq -r \
+          --arg IMAGE_NAME "$IMAGE_NAME" \
+          --arg TARGET_TAG "$TARGET_TAG" \
+          '
+              map(select(.RepoTags[] | contains($TARGET_TAG)))[0]
+              | .RepoDigests[]
+              | select(. | startswith($IMAGE_NAME))
+          '
+        )
```
dervoeti, line 88: I think `docker inspect "$TARGET_TAG" --format '{{index .RepoDigests 0}}'` would yield the same result in our case. But I'm fine with that variant as well since it's more strict, even though it's also more complex.

dervoeti (commented):
Minor nitpicks, LGTM overall!

.github/actions/publish-manifest/action.yml
NickLarsenNZ: Thanks for that. I spent longer than I'd like to admit trying to get that directly.

It is strange how `{{.Manifest}}` works without `println` (the default), yet `{{.Manifest.Digest}}` doesn't. In fact, it doesn't even error and uses the default. 🤯

.github/actions/publish-image/action.yml
NickLarsenNZ: Yeah, I just wanted the extra safety of getting the digest that matches, rather than using the first element.

dervoeti (approved):
LGTM

PR #811, opened 2024-08-21
fix(hbase): link to phoenix server jar

razvan:
# Description

A link is added to `/stackable/hbase/lib` that points to the Phoenix server jar file. It is assumed that that the jar file has the same version as the Phoenix source zip file.

This is not the case for Hbase 2.6 where the source file is called `phoenix-5.3.0-4afe457-src.tar.gz` but the artifact is called `phoenix-server-hbase-2.6-5.3.0-SNAPSHOT.jar` and it leads to `ClassNotFoundException` when trying to start the Phoenix server.

This fix changes the link to point to the more generic file that doesn't include the actual Phoenix version.

For the 2.4 images this is:

```shell
root@2051a2f8df93 /stackable/hbase-2.4.18 # ls -l /stackable/phoenix/
lrwxrwxrwx. 1 stackable stackable        34 Jun 11  2023 phoenix-server-hbase-2.4.jar -> phoenix-server-hbase-2.4-5.2.0.jar
...
```

While for the 2.6 images this is:

```shell
➜  ~ ls -l /stackable/phoenix/
lrwxrwxrwx. 1 razvan razvan        43 Jun 11  2023 phoenix-server-hbase-2.6.jar -> phoenix-server-hbase-2.6-5.3.0-SNAPSHOT.jar
...
```

```[tasklist]
```

<details>

```shell
```

</details>

### DIFF
--- a/hbase/Dockerfile
+++ b/hbase/Dockerfile
@@ -323,9 +323,9 @@ microdnf clean all
 rpm -qa --qf "%{NAME}-%{VERSION}-%{RELEASE}\n" | sort > /stackable/package_manifest.txt
 rm -rf /var/cache/yum
 
-ln -s /stackable/hbase-${PRODUCT} /stackable/hbase
-ln -s /stackable/hbase-operator-tools-${HBASE_OPERATOR_TOOLS} /stackable/hbase-operator-tools
-ln -s "/stackable/phoenix/phoenix-server-hbase-${HBASE_PROFILE}.jar" "/stackable/hbase/lib/phoenix-server-hbase-${HBASE_PROFILE}.jar"
+ln --symbolic --logical --verbose "/stackable/hbase-${PRODUCT}" /stackable/hbase
+ln --symbolic --logical --verbose "/stackable/hbase-operator-tools-${HBASE_OPERATOR_TOOLS}" /stackable/hbase-operator-tools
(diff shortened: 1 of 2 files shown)


hbase/Dockerfile
```diff
@@ -325,7 +325,7 @@ rm -rf /var/cache/yum
 
 ln -s /stackable/hbase-${PRODUCT} /stackable/hbase
 ln -s /stackable/hbase-operator-tools-${HBASE_OPERATOR_TOOLS} /stackable/hbase-operator-tools
-ln -s "/stackable/phoenix/phoenix-server-hbase-${HBASE_PROFILE}-${PHOENIX}.jar" "/stackable/hbase/lib/phoenix-server-hbase-${HBASE_PROFILE}-${PHOENIX}.jar"
```
sbernauer, line 328: I think ideally the `ln` command should fail when the target does not exist - this would have catched the error
I tested it locally and `--logical` seems like the thing we want, but I'm no `ln` expert
```
ln --verbose --logical foo bar
ln: failed to access 'foo': No such file or directory
```

razvan: Done https://github.com/stackabletech/docker-images/pull/811/commits/5ecdb565e7e9cac4293fa9193e346d374f4696ec

sbernauer (approved):
Thanks!

PR #812, opened 2024-08-21
fix: update ca-certificates version in the check to remove e-tugra cert

dervoeti:


stackable-base/Dockerfile
```diff
@@ -137,9 +137,8 @@ chown stackable:stackable /stackable/.profile
 
 # CVE-2023-37920: Remove "e-Tugra" root certificates
 # e-Tugra's root certificates were subject to an investigation prompted by reporting of security issues in their systems
-# The package ca-certificates 2023.07.22 fixes the problem, until ubi9-minimal updates to it, we should remove them
-# manually.
-if [ "$(rpm -qa ca-certificates)" != "ca-certificates-2023.2.60_v7.0.306-90.1.el9_2.noarch" ]; then
+# Until they are removed by default from ca-certificates, we should remove them manually
+if [ "$(rpm -qa ca-certificates)" != "ca-certificates-2024.2.69_v8.0.303-91.4.el9_4.noarch" ]; then
```
NickLarsenNZ, line 141: ca-certificates-2024.2.69_v8.0.303-91.4.el9_4.noarch still has the E-Tugra CAs in it (at least when I ran a docker build, they still appeared after running `trust list --filter=ca-anchors | grep 'E-Tugra'`

dervoeti: Yeah that's correct. We remove the E-Tugra certs by adding [blocklists](https://github.com/stackabletech/docker-images/tree/main/stackable-base/stackable/ca-cert-blocklist).
I think the check in the dockerfile is mainly a reminder for us to recheck whether the blocklists are still needed when `ca-certificates` is updated.
Maybe @sbernauer can verify this, I think he implemented it.

NickLarsenNZ: Ah ok.

It is unfortunate that builds bake hides the error so easily which causes a lot of wasted time.

Feel free to resolve, I'll submit a follow up if I can find a way to surface the error better.

sbernauer: Everything said makes sense :+1: I'm just a bit worried why `ca-certificates-2024.2` still did not remove the cert...

sbernauer (approved):
LGTM to me

stackable-base/Dockerfile
NickLarsenNZ: For posterity, the blocked certs that @dervoeti linked are copied over and double checked in:

https://github.com/stackabletech/docker-images/blob/ea720285770eb50c314661ad9fa5f47abaac99bc/stackable-base/Dockerfile#L150-L157

PR #814, opened 2024-08-27
Generate SBOMs during build

dervoeti:
Needed for https://github.com/stackabletech/issues/issues/614

This enables SBOM generation at build time.

Products where SBOM generation did not work yet:
 - Kafka: Problem with the Gradle CycloneDX plugin and Scala (CycloneDX/cyclonedx-gradle-plugin#239)
 - OPA / statsd_exporter: [cyclonedx-gomod](https://github.com/CycloneDX/cyclonedx-gomod) expects the source code to be in a Git repo and obtains the version from the Git tag. Solutions: a) Patch cyclonedx-gomod b) Create a fake Git repo c) Clone the source code from Git instead of mirroring it in Nexus

We can look into fixing that at a later point in time. The main value of build time SBOM generation is currently that it helps detecting how a dependency is brought into an image. Since Kafka and the Go-based products are not heavily affected and the fixes for both did not seem to be trivial, I figured that we could skip them for now since SBOM generation for these products won't provide too much value anyway at the moment. We can improve this later, we'll have to adapt SBOM generation continuously for all upcoming product versions anyway.

Another PR related to this one will follow, since this one only places an SBOM as a file inside the image. The other one will enable using this SBOM file to extend our actual published SBOMs with the dependency information (those who are attached to images and shown e.g. in our SBOM browser).

I tested all builds for LTS versions successfully on my local machine and verified that the SBOMs are present.


lfrancke:
I'd like to review this next week. If it has time I'd be grateful for that but this is no block or veto. I'll let you decide.

dervoeti:
I think the builds for product versions where I did not create a patch file that generates the SBOM would fail right now. We can discuss next week if we want to create patch files for those versions as well or make the "move SBOM" steps in the Dockerfiles optional ("only move it if the file exists").

hadoop/stackable/patches/3.4.0/005-cyclonedx-plugin.patch
```diff
@@ -0,0 +1,64 @@
...
++        </configuration>
++        <executions>
++          <execution>
++            <phase>package</phase>
++            <goals>
++              <goal>makeBom</goal>
++            </goals>
++          </execution>
++        </executions>
++      </plugin>
```
lfrancke, line 26: Can you check if this is really needed?

https://issues.apache.org/jira/browse/HADOOP-18590

Also: If it is needed, should it use the variable you defined later/earlier in the main pom.xml?

hive/Dockerfile
```diff
@@ -30,10 +30,12 @@ curl --fail -L "https://repo.stackable.tech/repository/packages/hive/apache-hive
 patches/apply_patches.sh ${PRODUCT}
 
 cd /stackable/apache-hive-${PRODUCT}-src/
+mvn --batch-mode --no-transfer-progress clean package -DskipTests --projects standalone-metastore
```
lfrancke, line 33: This seems wrong

airflow/Dockerfile
```diff
@@ -46,7 +46,8 @@ RUN python3 -m venv --system-site-packages /stackable/app && \
     pip install --no-cache-dir --upgrade pip && \
     pip install --no-cache-dir apache-airflow[${AIRFLOW_EXTRAS}]==${PRODUCT} --constraint /tmp/constraints.txt && \
     # Needed for pandas S3 integration to e.g. write and read csv and parquet files to/from S3
-    pip install --no-cache-dir s3fs
```
lfrancke, line 49: Can you check if this is really needed or if https://github.com/apache/airflow/pull/31835 helps?

trino/stackable/patches/451/001-cyclonedx-plugin.patch
```diff
@@ -0,0 +1,30 @@
+diff --git a/pom.xml b/pom.xml
```
lfrancke, line 1: We need this patch for 455 as well

airflow/Dockerfile
dervoeti: Their SBOM generation process is more complicated then ours, also their resulting SBOM lacks the dependency tree for Python packages. They do however contain NPM packages with their dependency relations, these are currently not included in the SBOM this PR generates. But Syft picks up the NPM components when scanning the image. We could look into generating an SBOM for the NPM components as well, the advantage would be the dependency tree information being present in our SBOMs. But this information is at the moment not really relevant for vulnerability analysis, so I think it's not worth the effort. We can always improve it later if we want to. For analysis of vulnerabilities the Python package dependency tree is useful, which is generated by this PR.

hive/Dockerfile
dervoeti: Fixed

hadoop/stackable/patches/3.4.0/005-cyclonedx-plugin.patch
dervoeti: It was not needed, I removed it

trino/stackable/patches/451/001-cyclonedx-plugin.patch
dervoeti: Done

dervoeti:
@lfrancke I fixed the issues mentioned in your comments and added patches for all product versions we currently build.

lfrancke (approved):
It looks good to me but I did not test it myself

PR #803, opened 2024-08-13
fix: updated versions.py files to use vector 0.40.0

dervoeti:
Related to https://github.com/stackabletech/docker-images/pull/802, fixes the builds by pointing to the new Vector version

PR #815, opened 2024-08-27
chore(vector): Bump to 0.40.1

NickLarsenNZ:
# Description

Replaces #802 and #803.

> _Fixes a Vector v0.40.0 regression where the reduce transform would not group top level objects correctly._

_container build works_

### DIFF
--- a/airflow/versions.py
+++ b/airflow/versions.py
@@ -5,7 +5,7 @@ versions = [
         "git_sync": "v4.2.4",
         "statsd_exporter": "0.26.1",
         "tini": "0.19.0",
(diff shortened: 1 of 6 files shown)


sbernauer (approved):
Normally we would wait for the patch level bump meeting, but you did the work already ;)

NickLarsenNZ:
> Normally we would wait for the patch level bump meeting, but you did the work already ;)

It was purposeful:

image

Seemed pertinent that log tests are based on this fix.

I mentioned on Slack that it would have otherwise been left until we fix versions.

PR #824, opened 2024-09-05
fix(ci): Install hadolint in pre-commit workflow

Techassi:
Part of https://github.com/stackabletech/issues/issues/637

This fixes two things:

- Install the hadolint binary to be able to run the pre-commit hook
- Set the default language version of `node` to `system`. This fixes issues on NixOS systems.

### DIFF
--- a/.github/workflows/pr_pre-commit.yaml
+++ b/.github/workflows/pr_pre-commit.yaml
@@ -22,12 +22,17 @@ jobs:
         run: |
           set -xeuo pipefail
 
-          LOCATION="$HOME/hadolint"
+          LOCATION_DIR="$HOME/.local/bin"
+          LOCATION_BIN="$LOCATION_DIR/hadolint"
+
           SYSTEM=$(uname -s)
           ARCH=$(uname -m)
 
-          curl -sL -o "${LOCATION}" "https://github.com/hadolint/hadolint/releases/download/${{ env.HADOLINT_VERSION }}/hadolint-$SYSTEM-$ARCH"
-          chmod 700 "${LOCATION}"
+          mkdir -p "$LOCATION_DIR"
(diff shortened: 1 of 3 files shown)


.github/workflows/pr_pre-commit.yaml
```diff
@@ -14,6 +17,17 @@ jobs:
       - uses: actions/setup-python@82c7e631bb3cdc910f68e0081d67478d79c6982d # v5.1.0
         with:
           python-version: '3.12'
+      - name: Setup Hadolint
+        shell: bash
+        run: |
+          set -xeuo pipefail
+
+          LOCATION="$HOME/hadolint"
```
NickLarsenNZ, line 25: Does it need adding to the GITHUB_PATH?

NickLarsenNZ (commented):
I'm not sure $HOME is in the path.

.github/workflows/pr_pre-commit.yaml
Techassi: Yeah that's the thing I'm not sure about. I didn't find any information on that in their documentation.

Techassi: Indeed, `$HOME` didn't work, we need to use `$GITHUB_PATH`.

.github/workflows/pr_pre-commit.yaml
```diff
@@ -14,6 +17,22 @@ jobs:
       - uses: actions/setup-python@82c7e631bb3cdc910f68e0081d67478d79c6982d # v5.1.0
         with:
           python-version: '3.12'
+      - name: Setup Hadolint
+        shell: bash
+        run: |
+          set -xeuo pipefail
```
NickLarsenNZ, line 23: We can remove debug now, if you want. I don't mind if it stays in though.

```suggestion
          set -euo pipefail
```

NickLarsenNZ (approved):
LGTM

NickLarsenNZ (approved):
LGTM

PR #828, opened 2024-09-06
Fixes for SBOM build time generation

dervoeti:
Fix for https://github.com/stackabletech/docker-images/pull/814

This includes missing patches for Omid, trino-storage, hbase-operator-tools and Phoenix, which currently cause some builds to break.


lfrancke (commented):
This does not contain anything to copy the generated SBOMs. Is that intended?

dervoeti:
Yes, the copy/move commands are already present in the Dockerfiles. The last PR also contained patches for these products, but not for certain versions. The patches for these other versions are contained in this PR.

lfrancke (approved):

PR #841, opened 2024-09-10
fix: jq arg parsing

dervoeti:
The args to `jq` were not parsed correctly, which resulted in e.g.
```
...
"purl": "$purl",
...
```
in the final JSON of the SBOM.

I could not figure out how to concatenate the two args with `jq` (`$name` and `$description`) so I concatenated both variables with Shellscript into one arg for `jq`.

### DIFF
--- a/.github/actions/publish-image/action.yml
+++ b/.github/actions/publish-image/action.yml
@@ -129,19 +129,18 @@ runs:
         # TODO (@Techassi): Replace author with manufacturer, because it is
         # automated, see https://cyclonedx.org/docs/1.6/json/#metadata_component_manufacturer
         jq -s \
-          --arg description "$IMAGE_METADATA_DESCRIPTION" \
-          --arg name "$IMAGE_METADATA_NAME" \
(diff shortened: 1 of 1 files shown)


razvan:
> I could not figure out how to concatenate the two args with `jq` (`$name` and `$description`) so I concatenated both variables with Shellscript into one arg for `jq`.

Have you tried [join()](https://jqlang.github.io/jq/manual/#join) ?

dervoeti:
> > I could not figure out how to concatenate the two args with `jq` (`$name` and `$description`) so I concatenated both variables with Shellscript into one arg for `jq`.
> 
> Have you tried [join()](https://jqlang.github.io/jq/manual/#join) ?

Yeah, couldn't make it work. But I didn't try too much since joining the string with Shellscript works fine and shouldn't be a problem.

NickLarsenNZ:
I have concatenated via `\($a\) \($b\)`, eg:

```
❯ jq -n --arg a hello --arg b world '"\($a) \($b)"'
"hello world"
```

NickLarsenNZ (approved):
LGTM. Left another comment on how to concat with `jq` for future reference.

PR #843, opened 2024-09-10
fix: Don't use release profile for phoenix build

dervoeti:
In [this PR](https://github.com/stackabletech/docker-images/pull/814) I enabled the CycloneDX plugin for the `release` profile of Phoenix and changed the Maven command to build Phoenix using that profile (because that's how HBase does it as well).

However, the Phoenix `release` profile also enables the [apache-rat-plugin](https://creadur.apache.org/rat/apache-rat-plugin/) which causes a minor problem with our current build of Phoenix 5.3.0.

Since using the release profile does not seem to have any further advantages I moved the CycloneDX plugin to the general plugin section in the pom.xml and do not activate the release profile anymore when building Phoenix.

### DIFF
--- a/hbase/Dockerfile
+++ b/hbase/Dockerfile
@@ -248,7 +248,6 @@ mvn \
   -Dhadoop.version=${HADOOP} \
   -DskipTests \
   -Dcheckstyle.skip=true \
-  -Prelease \
   -fphoenix-${PHOENIX}-src \
   clean \
   package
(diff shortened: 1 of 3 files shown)


razvan (approved):
I just scanned the pom changes briefly but didn't test them. I trust you.

Maybe a less invasive way would be to define our own CycloneDx profile that we enable per command line in addition to the default one. Just an idea.

PR #849, opened 2024-09-11
Make uid/gid configurable & change group of files

lfrancke:
# Description

This PR contains three related changes for UID/GID handling. I decided to lump these together because they are related but if whoever reviews this feels more comfortable I can also split this in multiple PRs.

- Make user id & group configurable
- This also switches to use numeric UIDs everywhere instead of hardcoding a name (e.g. `stackable`)
- Change ownership of all things belonging to `stackable` so that they are owned by the `root` group (gid = 0)

## Configurable user name, uid and gid

Using the new [functionality](https://github.com/stackabletech/image-tools/pull/41) to support global arguments this extracts the user id, user name and gid into arguments that can be changed easily.
They still stay at the `1000` we use so far even though that is not optimal and needs to be changed as well.

But because I don't know if any operators make any assumptions about the uid/gid (and fsgroup which is not handled here) I decided to split this into two steps.

This PR is step 1: Make things more configurable, step 2 will follow later.

### Detailed reasoning

Using a hardcoded uid for our `stackable` user is a good idea in theory, in practice the id 1000 should be avoided.

This is because the users from Docker containers are mapped to users on the underlying host OS. Some OSes start "real" user ids at 1000 (or 500) and reserve everything before that to "system" users. User 1000 therefore has a good chance of being mapped to a real user that exists on the underlying system, which should be avoided.

The easiest way of doing so is by picking an arbitrarily large (more or less) number to statically use in our Dockerfiles.
This is exactly what OpenShift does by default. It picks a "random" UID from a range of UIDs (in reality it picks the first one from a range). The UID is larger than 1.000.000.000 by default.

>[!NOTE]
> There is a [bug](https://github.com/moby/moby/issues/5419)/problem and the number cannot be too large. The linked issue does include a workaround which [did apply](https://github.com/stackabletech/docker-images/pull/781) to our base images.

>[!NOTE]
> Kubernetes 1.30 contains _[user namespaces](https://github.com/kubernetes/enhancements/blob/master/keps/sig-node/127-user-namespaces/README.md)_ as a beta feature. Expected to be moved to GA at a later point but at the moment not for 1.32 so the earliest would be 1.33 (around April/May 2025) and it would be another 1,5-2 years before we could use it.

## Numeric UIDs

> [!NOTE]
> This is to support `securityContext.runAsNonRoot` for users wanting to use it or in preparation for the future where we might want to enable it ourselves

The `USER` statement in a Dockerfile ends up in an image's metadata:

cfefc85f-cad4-4042-b3cb-21a31629e304

This user is used as the default user when an image is started using plain Docker:

```bash
docker run -it --entrypoint bash docker.stackable.tech/stackable/druid:30.0.0-stackable0.0.0-dev
```

It is also the default when used as a plain Pod in Kubernetes:

```bash
kubectl run test --image=docker.stackable.tech/stackable/druid:30.0.0-stackable0.0.0-dev --rm=true --restart=Never --tty=true --stdin=true -- bash
```

In OpenShift this is what it looks like as an admin user (they are exempt from SCCs):

```bash
kubectl run test --image=docker.stackable.tech/stackable/hbase:2.6.0-stackable0.0.0-dev --rm=true --restart=Never --tty=true --stdin=true --namespace test -- id  

Warning: would violate PodSecurity "restricted:v1.24": allowPrivilegeEscalation != false (container "test" must set securityContext.allowPrivilegeEscalation=false), unrestricted capabilities (container "test" must set securityContext.capabilities.drop=["ALL"]), runAsNonRoot != true (pod or container "test" must set securityContext.runAsNonRoot=true), seccompProfile (pod or container "test" must set securityContext.seccompProfile.type to "RuntimeDefault" or "Localhost")
uid=1000(stackable) gid=1000(stackable) groups=1000(stackable)
```

Here is the same command run as a non-admin user (note the use of a non-1000 ID means that we bypass the SCC warning):

```bash
oc run test --as developer --image=docker.stackable.tech/stackable/hbase:2.6.0-stackable0.0.0-dev --rm=true --restart=Never --tty=true --stdin=true --namespace test -- id
uid=1000740000(1000740000) gid=0(root) groups=0(root),1000740000
pod "test" deleted
```

If we - or someone else - want to enforce that a user is non-root using the `securityContext.runAsNonRoot` field it will not work as Kubernetes has no way of mapping the string `stackable` to a UID (it is not aware of the implementation details inside the container, it could call out to LDAP for all it knows). Therefore this combination (non-numeric UID) and `runAsNonRoot` is forbidden and results in an error:

a017e819-3866-4f49-902e-17688147e100

This PR, therefore, switches all Dockerfiles to use the numeric UID instead of the username.

## Group of all files
To support our images to run as an arbitrary user we need to make sure that arbitrary users can read, write and execute all files and commands that user stackable can as well.

The container user is always a member of the root group and we're applying the [suggested](https://docs.openshift.com/container-platform/4.16/openshift_images/create-images.html#use-uid_create-images) steps.

```[tasklist]
```

<details>

```shell
```

</details>


stackable-base/Dockerfile
```diff
@@ -156,8 +159,8 @@ if [ "$(trust list --filter=ca-anchors | grep -c 'E-Tugra')" != "0" ]; then
 fi
 EOF
 
-COPY --from=product-utils-builder --chown=stackable:stackable /config-utils/target/release/config-utils /stackable/config-utils
-COPY --from=product-utils-builder --chown=stackable:stackable /config-utils/config-utils.cdx.xml /stackable/config-utils.cdx.xml
+COPY --from=product-utils-builder --chown=${STACKABLE_USER_UID}:0 /config-utils/target/release/config-utils /stackable/config-utils
```
razvan, line 162: If I understand the linked docs correctly, the group permissions must also be the same as the user permissions. This might be implicit here (I don't know) but we might want to also be explicit about it.

razvan (commented):
so far, so good :)

Techassi (commented):
I know this is currently still a draft, but I wanted to make sure that the correct image-tools version is used here: https://github.com/stackabletech/docker-images/blob/main/.github/actions/build-product-image/action.yml#L11-L13

lfrancke:
Thanks @Techassi. Good catch. I'll update it now.

stackable-base/Dockerfile
lfrancke: Yeah, we set the group to `0` as per the recommendation (and it makes sense as well).
I'm not sure if I understand you Razvan. Could you rephrase your comment for me?

razvan: I mean the actual permissions have to be the same for both entities, not 755 for user and 000 for the group (for example)

lfrancke: That is what this line should be doing

```
chmod -R g=u /stackable
```

https://github.com/stackabletech/docker-images/pull/849/files#diff-d27bbada86d18407f7107257ec59dcaf6375b084613d33ab71d3499c3fa84441R113

Would this cover your concern?

razvan: Yes, I hadn't seen that.

lfrancke:
This is ready for review.
I did _not_ run any tests yet but I'll try to to that this week.

hadoop/Dockerfile
```diff
@@ -8,12 +8,12 @@ ARG JMX_EXPORTER
 ARG PROTOBUF
 ARG TARGETARCH
 ARG TARGETOS
+ARG STACKABLE_USER_UID
```
NickLarsenNZ, line 11: Just a thought, should these be defaulting to something so that they don't need specifying when doing a local build?

Nothing to change right now, just wanted to ask.

lfrancke: They should default to the current `1000`

See here: https://github.com/stackabletech/docker-images/pull/849/files#diff-6e03d3d89630aba3403afea3514a3ac5948e70d1ced091012e56446f614da340

Or do you mean something different?

lfrancke: I see that I also put it in the PR description actually. I believe that works if you use image_tools 0.0.13

> Using the new https://github.com/stackabletech/image-tools/pull/41 to support global arguments this extracts the user id, user name and gid into arguments that can be changed easily.
They still stay at the 1000 we use so far even though that is not optimal and needs to be changed as well.

hello-world/Dockerfile
```diff
@@ -13,13 +14,20 @@ LABEL name="Hello World" \
       summary="The Stackable image for the Stackable Hello World Operator" \
       description="This image is deployed by the Stackable Hello World Operator."
 
-RUN microdnf update && \
-    microdnf clean all && \
-    rm -rf /var/cache/yum
+RUN <<EOF
+microdnf update
+microdnf clean all
+rm -rf /var/cache/yum
```
NickLarsenNZ, line 21: It looks like some images skip this step. Should be be done on all, or it is only needed here?

Nothing to change here, just wanted to ask.

lfrancke: This should be done everywhere and I thought I checked that they did (at least the ones this PR is touching) but I'll check again, thanks!

NickLarsenNZ (dismissed):
LGTM

lfrancke:
Thank you! I'll let the tests for all operators run before I merge.

hadoop/Dockerfile
NickLarsenNZ: > They should default to the current 1000

The default to `1000` in bake, but not in the Dockerfile (the ARG has no default).

lfrancke: That is by intention, yes. The whole idea here was to only specify the UID in a single place so we can easily update it once we are ready to do so.

lfrancke:
Tests:

- Airflow: not done yet
- Druid: not done yet
- Hadoop: not done yet
- HBase: not done yet
- Hello World: not done yet
- Hive:not done yet

lfrancke:
This is now not ready to merge anymore because it first needs https://github.com/stackabletech/actions/pull/2 to be merged and then an update to the actions.

lfrancke:
All tests pass. I now just need to update the action.

lfrancke:
Action has been updated. This is ready for review again.

hadoop/Dockerfile
NickLarsenNZ: I'm meaning the Dockerfile should have defaults (bake overrides those defaults), but since we don't do that for any other arg, we can ignore it.

hello-world/Dockerfile
lfrancke: I checked. All the ones I touch have it.

NickLarsenNZ: I did a quick check based on counting lines, and see there is a difference:

```
❯ rg --color=always 'microdnf update' | wc -l
42

❯ rg --color=always 'microdnf clean' | wc -l
38

❯ rg --color=always '/var/cache/yum' | wc -l
37
```

stackable-base/Dockerfile
```diff
@@ -110,21 +113,22 @@ microdnf install \
...
 # Making this a system user prevents a mail dir from being created, expiry of passwords etc. but it will warn:
 #   useradd warning: stackable's uid 1000 is greater than SYS_UID_MAX 999
 # We can safely ignore this warning, to get rid of the warning we could change /etc/login.defs but that does not seem worth it
+# We'll leave the home directory hardcoded to /stackable because I don't want to deal with which chars might be valid and which might not in user name vs. directory
 useradd \
   --no-log-init \
-  --gid stackable \
-  --uid 1000 \
+  --gid ${STACKABLE_USER_GID} \
+  --uid ${STACKABLE_USER_UID} \
```
NickLarsenNZ, line 127: And issue for later: quote variable expensions (would be picked up by shellcheck)

NickLarsenNZ (approved):
LGTM

lfrancke:
Followup PR with more products: https://github.com/stackabletech/docker-images/pull/890

lfrancke:
# Release Notes

- Our Docker images now exclusively make use of numeric user IDs in `USER` statements allowing the use of `securityContext.runAsNonRoot`
- The group id of all files relevant to our products is now set to `0`. This allows the images to be used with any arbitrary user as every container user will always belong to the root group (`0`). This is especially useful on OpenShift when trying to move to the `restricted-v2` SecurityContextConstraint (SCC), Stackable currently defaults to the `nonroot-v2` SCC but we plan on migrating to `restricted-v2` in the future

Issue #847, opened 2024-09-11
chore(airflow): Update container images ahead of Stackable Release 24.11


NickLarsenNZ:
Part of #831.

- Add: `2.10.2` (`2.9.3` remains the latest/LTS, and was already added in #809).
- Remove: `2.6.x`, `2.8.x` (already done in #809).

> [!TIP]
> Please add the `scheduled-for/20XX-XX` label.

```[tasklist]
### Update tasks
- [x] Update `versions.py` to reflect the agreed upon versions in the spreadsheet (including the removal of old versions).
- [x] Download new constraints file (see `airflow/download_constraints.sh`).
- [x] Update other dependencies if applicable (eg: python, statsd_exporter, etc).
- [ ] Check other operators (getting_started / kuttl) for usage of the versions. Add the PR(s) to the list below.
- [ ] Update the version in demos. Add the PR(s) to the list below.
- [ ] Research memory OOM problems in demo when using Airflow 2.10.2 -> Maybe increase default memory for webservers
```

```[tasklist]
### Related Pull Requests
- [ ] https://github.com/stackabletech/docker-images/pull/877
- [ ] https://github.com/stackabletech/airflow-operator/pull/512
```

```[tasklist]
### Acceptance
- [x] Can build image locally
- [x] Kuttl smoke tests passes locally
- [ ] Release notes written in a comment below
- [ ] Applicable `release-note` label added to this issue
```

<details>
<summary>Testing instructions</summary>

```shell
# See the latest version at https://pypi.org/project/image-tools-stackabletech/
pip install image-tools-stackabletech==0.0.12

bake --product airflow=x.y.z # where x.y.z is the new version added in this PR

kind load docker-image docker.stackable.tech/stackable/airflow:x.y.z-stackable0.0.0-dev

# Change directory into the airflow-operator repository and update the
# product version in tests/test-definition.yaml
./scripts/run-tests --test-suite smoke-latest # or similar
```

</details>

_Please consider updating this template if these instructions are wrong, or
could be made clearer._

NickLarsenNZ:
I see a new statsd_exporter version available. Will update that in a separate PR.

NickLarsenNZ:
Smoke tests pass:

```
--- PASS: kuttl (423.52s)
    --- PASS: kuttl/harness (0.00s)
        --- PASS: kuttl/harness/smoke_airflow-2.10.2_openshift-false_executor-kubernetes (423.49s)
PASS
```

[...]

PR #877, opened 2024-09-23
feat: Add airflow 2.10.2

NickLarsenNZ:
# Description

Part of https://github.com/stackabletech/docker-images/issues/847

- Add Airflow 2.10.2 (Python 3.12)
  - Improve the download constraints script.
  - Explicitly install the python package.
  - Use python at the specific version - because there is no alias for `python` or `python3`.
- Ignore hadlint check for pinning pip packages.
  - This can be its own piece of work.

The image builds pass for all versions.


sbernauer (approved):
Nice to see Python 3.12

PR #889, opened 2024-10-04
Ignore Docker warning

lfrancke:
# Description

It'd warn about GIT_SYNC being possibly undefined. This is correct in theory, in practice however we always set it from the outside and defining it here could lead to version skew or another place to update versions.


NickLarsenNZ (commented):
Relates to https://github.com/stackabletech/docker-images/pull/849#pullrequestreview-2340472292

I think we should set defaults in Dockerfiles (we just happen to set them in bake, but IMO we shouldn't assume bake is used for Dockerfiles (yes, I know many of our builds depend on layers having already been built locally, but decoupling where we can would be nice).

NickLarsenNZ (approved):

PR #890, opened 2024-10-08
Make uid/gid configurable & change group of files - part 2

lfrancke:
# Description

Make uid/gid configurable & change group of files.
This is a follow-up for https://github.com/stackabletech/docker-images/pull/849 (please check there for details).

This includes ( :heavy_check_mark:   means all tests have been run)

- The missing bits for Hive :heavy_check_mark:  
- Kafka :heavy_check_mark:  (fixed https://github.com/stackabletech/kafka-operator/pull/766 while doing this)
- (Kafka Testing Tools)
- (KCat)
- NiFi :heavy_check_mark:  (fixed https://github.com/stackabletech/nifi-operator/pull/687 while doing this)
- Omid :heavy_check_mark: 
- OPA :heavy_check_mark: (implemented https://github.com/stackabletech/opa-operator/pull/633 while doing this)
- Spark :heavy_check_mark: 

There are still a few tools missing and they'll come in a follow-up PR

```[tasklist]
```

<details>

```shell
```

</details>


CHANGELOG.md
```diff
@@ -31,6 +31,7 @@ All notable changes to this project will be documented in this file.
 - Enable [Docker build checks](https://docs.docker.com/build/checks/) ([#872]).
 - java: migrate to temurin jdk/jre ([#894]).
 - tools: bump kubectl to `1.31.1` and jq to `1.7.1` ([#896]).
+- Make username, user id, group id configurable, use numeric ids everywhere, change group of all files to 0 ([#849], [#890])
```
NickLarsenNZ, line 34: oops, I forgot to add the `.` during the conflict resolution :grimacing: 

```suggestion
- Make username, user id, group id configurable, use numeric ids everywhere, change group of all files to 0 ([#849], [#890]).
```

hive/Dockerfile
```diff
@@ -103,40 +103,47 @@ LABEL io.openshift.tags="ubi9,stackable,hive,sdp"
...
-RUN cp /stackable/hadoop/share/hadoop/tools/lib/azure-storage-${AZURE_STORAGE}.jar /stackable/hive-metastore/lib/
-RUN cp /stackable/hadoop/share/hadoop/tools/lib/azure-keyvault-core-${AZURE_KEYVAULT_CORE}.jar /stackable/hive-metastore/lib/
+cp /stackable/hadoop/share/hadoop/tools/lib/hadoop-azure-${HADOOP}.jar /stackable/hive-metastore/lib/
+cp /stackable/hadoop/share/hadoop/tools/lib/azure-storage-${AZURE_STORAGE}.jar /stackable/hive-metastore/lib/
+cp /stackable/hadoop/share/hadoop/tools/lib/azure-keyvault-core-${AZURE_KEYVAULT_CORE}.jar /stackable/hive-metastore/lib/
+
+# All files and folders owned by root to support running as arbitrary users
+# This is best practice as all container users will belong to the root group (0)
+chown -R ${STACKABLE_USER_UID}:0 /stackable
+chmod -R g=u /stackable
```
NickLarsenNZ, line 139: Doesn't this make files and folders owned by STACKABLE_USER_ID? It sets the group to the root group (and then sets the group permissions to be the same as the owner).

Maybe I misinterpreted.

_I won't leave the same comment for other occurrences._

opa/Dockerfile
```diff
@@ -85,32 +85,41 @@ FROM stackable/image/vector
...
-COPY --from=multilog-builder --chown=stackable:stackable /daemontools/admin/daemontools/command/multilog /stackable/multilog
+COPY --chown=${STACKABLE_USER_UID}:0 opa/stackable/bin /stackable/opa/bin
 
-COPY --chown=stackable:stackable opa/stackable/bin /stackable/opa/bin
+RUN <<EOF
+microdnf update
+
+# jq: Required for filtering logs
+microdnf install \
+  jq
```
NickLarsenNZ, line 111: Just noting that we install jq in the tools image. I'm unsure where that is used.

NickLarsenNZ (approved):
LGTM, I assume the images build locally.

hive/Dockerfile
lfrancke: Yes, that's the intention. To me that looks correct. 

I lifted it from the docs:
https://docs.openshift.com/container-platform/4.16/openshift_images/create-images.html#use-uid_create-images

Maybe you can take a quick look at it and see if what I do matches the intention. That said: The final images look good so I'm fairly certain.

opa/Dockerfile
lfrancke: I'm not sure I understand. Can you explain that to me? Is the tools image used here?

hive/Dockerfile
NickLarsenNZ: The group is being set to root, but the comment says the owner is set to root. To me, that is the incorrect part.

```
chown $OWNER:$GROUP
```
```

opa/Dockerfile
NickLarsenNZ: What is the intention of the tools image? I assumed it was so we could use `COPY` to pull out tools for use, rather than installing them via microdnf. But maybe it is for a totally different use case.

Or, if we install `jq` via microdnf, why don't we do the same for tools?

https://github.com/stackabletech/docker-images/blob/d3231ba9d6291c7c1f31404e3e2c472d433f3042/tools/Dockerfile#L41-L43

lfrancke: Ah.... I don't actually know....sorry.

NickLarsenNZ: No problem. I will find out separately from this PR.

hive/Dockerfile
lfrancke: Ahhhhh! The comment should just say "root group"

> For an image to support running as an arbitrary user, directories and files that are written to by processes in the image must be owned by the root group and be read/writable by that group. Files to be executed must also have group execute permissions.

Good catch! I'll update this everywhere in this PR.

lfrancke:
And to answer your question: Yes, it builds locally _and_ I did run _all_ tests (see the issue description) and it all looks good.

hive/Dockerfile
lfrancke: Done in [`4411137` (#890)](https://github.com/stackabletech/docker-images/pull/890/commits/4411137fa9d043c3d77d40593b47c7dce949bfbd)

NickLarsenNZ (approved):
LGTM

PR #893, opened 2024-10-10
chore: Remove unused scripts

Techassi:
This removes a bunch of unused scripts.

Some of these scripts need to be added to https://github.com/stackabletech/actions and referred to with `${{ github.action_path }}` in these actions. All files required to run these actions must be contained in the `stackabletech/actions` repository.


lfrancke:
Marked as "draft" but is in "Waitig for review"? Shall we move it back for now?

xeniape:
The `enumerate-product-versions.py` script seems to still be used in the [release workflow](https://github.com/stackabletech/docker-images/blob/8ea476e07e7516dd5ce1f7c4ab19f5f4ce65e743/.github/workflows/release.yml#L190)

NickLarsenNZ (changes requested):
As per @xeniape's comment, we should only remove the `enumerate-product-versions.py` scripts once it is no longer used.

NickLarsenNZ (approved):
LGTM

PR #898, opened 2024-10-17
feat(hbase): add hbase-entrypoint.sh script

razvan:
# Description

prere-quisite for : https://github.com/stackabletech/hbase-operator/pull/570

```[tasklist]
```

<details>

```shell
```

</details>


hbase/stackable/bin/hbase-entrypoint.sh
```diff
@@ -0,0 +1,72 @@
...
+}
+
+handle_term_signal() {
+  if [ "${term_child_pid}" ]; then
+    if [ "regionserver" == "${HBASE_ROLE_NAME}" ]; then
+      echo "Start pre-shutdown"
+      # REGION_MOVER_OPTS is a string that contains multiple arguments and needs to be spliced here
+      # therefore disable shellcheck for this line
+      # shellcheck disable=SC2086
+      /stackable/hbase/bin/hbase org.apache.hadoop.hbase.util.RegionMover ${REGION_MOVER_OPTS}
```
NickLarsenNZ, line 38: There is a valid way to do it without disabling the shell check, but I can't recall it right now.

Nothing to do here though, if I find it I will mention it.

hbase/stackable/bin/hbase-entrypoint.sh
```diff
@@ -0,0 +1,72 @@
...
+}
+
+wait_for_termination() {
+  set +e
+  term_child_pid=$1
+  if [[ -v term_kill_needed ]]; then
+    kill -TERM "${term_child_pid}" 2>/dev/null
+  fi
+  wait "${term_child_pid}" 2>/dev/null
+  trap - TERM
```
NickLarsenNZ, line 54: What does this do? Run nothing on SIGTERM?

NickLarsenNZ (approved):
LGTM

hbase/stackable/bin/hbase-entrypoint.sh
razvan: [Copy&pasta from op-rs.](https://github.com/stackabletech/operator-rs/blob/82a0ef6d14c6f4e144ee1b4f5eff6abebf0108f2/crates/stackable-operator/src/utils/bash.rs#L40) 

I don't quite get understand it either.

NickLarsenNZ (approved):
Approving so we aren't blocked, but would like some light docs on the traps/signal handlin.

lfrancke:
@razvan would we revert this if we decide not to merge the other related graceful shutdown tickets?

razvan:
It's not used and has no impact if it stays.

Personally I would prefer to create entry point scripts for all products and get rid of the hard-coded shell snippets in the operators.

But if it bothers you, I can revert it.

lfrancke:
It doesn't. I was just wondering.

PR #307, opened 2023-02-03
Fix broken Superset image 23.4.0-rc1

siegfriedweber:
setuptools pinned to version 66.1.1.

The setuptools version 67.0.0 and newer are not compatible with the supported Superset versions because parsing of the requirements is more strict and fails in Superset (see also pypa/setuptools#3790).

This fixes the following error when starting Superset:

```
Traceback (most recent call last):
  File "/stackable/app/lib64/python3.8/site-packages/pkg_resources/__init__.py", line 3030, in _dep_map
    return self.__dep_map
  File "/stackable/app/lib64/python3.8/site-packages/pkg_resources/__init__.py", line 2826, in __getattr__
    raise AttributeError(attr)
AttributeError: _DistInfoDistribution__dep_map

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "/stackable/app/lib64/python3.8/site-packages/pkg_resources/_vendor/packaging/requirements.py", line 35, in __init__
    parsed = parse_requirement(requirement_string)
  File "/stackable/app/lib64/python3.8/site-packages/pkg_resources/_vendor/packaging/_parser.py", line 64, in parse_requirement
    return _parse_requirement(Tokenizer(source, rules=DEFAULT_RULES))
  File "/stackable/app/lib64/python3.8/site-packages/pkg_resources/_vendor/packaging/_parser.py", line 82, in _parse_requirement
    url, specifier, marker = _parse_requirement_details(tokenizer)
  File "/stackable/app/lib64/python3.8/site-packages/pkg_resources/_vendor/packaging/_parser.py", line 120, in _parse_requirement_details
    specifier = _parse_specifier(tokenizer)
  File "/stackable/app/lib64/python3.8/site-packages/pkg_resources/_vendor/packaging/_parser.py", line 209, in _parse_specifier
    tokenizer.consume("WS")
  File "/usr/lib64/python3.8/contextlib.py", line 120, in __exit__
    next(self.gen)
  File "/stackable/app/lib64/python3.8/site-packages/pkg_resources/_vendor/packaging/_tokenizer.py", line 183, in enclosing_tokens
    self.raise_syntax_error(
  File "/stackable/app/lib64/python3.8/site-packages/pkg_resources/_vendor/packaging/_tokenizer.py", line 163, in raise_syntax_error
    raise ParserSyntaxError(
pkg_resources.extern.packaging._tokenizer.ParserSyntaxError: Expected closing RIGHT_PARENTHESIS
    pytz (>dev)
[...]

PR #900, opened 2024-10-21
chore: update setuptools

dervoeti:
# Description

Setuptools was pinned to version 66.1.1 in https://github.com/stackabletech/docker-images/pull/307. That package contains a vulnerability (CVE-2024-6345). I bumped the dependency to the latest version, building and running Superset in my test cluster worked fine. I only tested with Superset 4.0.2, as that is currently the only version we plan to ship with SDP 24.11.

### DIFF
--- a/superset/Dockerfile
+++ b/superset/Dockerfile
@@ -45,7 +45,7 @@ RUN python3 -m venv /stackable/app \
     && pip install \
         --no-cache-dir \
         --upgrade \
-        setuptools==66.1.1 \
+        setuptools==75.2.0 \
         pip \
     && pip install \
         --no-cache-dir \


lfrancke:
Does setuptools end up in the final image? (Doesn't change the fact that we should upgrade)

dervoeti:
It does, you can see the dependency tree here for example: https://secobserve.stackable.tech/#/observations/1514479/show

I have not validated or analyzed it further though. I don't know too much about Python, but I also thought that `setuptools` was not a runtime dependency.

lfrancke:
I also don't know enough about Python but to me that's unexpected. But what do I know :)
I'm happy to approve this but you requested Sigi...I'll approve it anyway. Up to you to wait or not.

lfrancke (approved):

PR #915, opened 2024-11-05
ci: Add failure notifications to workflows

NickLarsenNZ:
# Description

Add a job at the end of each workflow which only runs on failure to send a slack notitification.

<details>
<summary>Considerations</summary>

We had the following four options to choose from:

1. Manually create threads per workflow, and hard code thread ids into each workflow.
    - Continuous log of build successes/failures along with relevant conversations.
    - Main thread stays clean.
2. _Be smat_ and get a thread-id from a cached artifact, and if it doesn't exist, create a new one.
    - Same as the first option, but without all the manual work. Good for adding new workflows.
3. Create a thread per workflow run (begin workflow, build, manifest, done)
4. Only message on failure (higher signal-to-noise, people might be happy to receive alerts for it since each would require action)

We opted for _Option 4_ to keep things simple and to increase signal-to-noise (alerts are actionable).

We have a further idea of having each [action](https://github.com/stackabletech/actions) write some high-level results to a file/output, and have a wrapping action that collates them all and builds the slack payload. This will be a little complicated with matrix jobs.

</details>

### DIFF
--- a/.github/workflows/build_airflow.yaml
+++ b/.github/workflows/build_airflow.yaml
@@ -114,3 +114,49 @@ jobs:
           image-registry-password: ${{ secrets.HARBOR_ROBOT_SDP_GITHUB_ACTION_BUILD_SECRET }}
           image-repository: sdp/${{ env.PRODUCT_NAME }}
           image-index-manifest-tag: ${{ matrix.versions }}-stackable0.0.0-dev
+
+  notify:
+    name: Failure Notification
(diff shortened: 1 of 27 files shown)


soenkeliebau (approved):
lgtm - full disclosure, have only skimmed the code, no way whatsoever of testing it :)

Techassi (approved):

PR #916, opened 2024-11-05
Change default user id and gid

lfrancke:
# Description
_Part of https://github.com/stackabletech/issues/issues/645_

Change default user id and gid to the same ones we use in the operators themselves.
These are only the defaults that are used when a Pod does not specify their own `securityContext.runAsUser` or `runAsGroup`.

As of now all our operators do set these (and `fsGroup`) to 1000 and 0 respectively.
As the next step we want to remove that hardcoding so the default would then fall back to what we specify here.

Therefore I do believe that this PR should be a simple change with no downstream consequences until the PRs from https://github.com/stackabletech/issues/issues/651 are merged.

<details>

```shell
```

</details>

### DIFF
--- a/conf.py
+++ b/conf.py
@@ -98,7 +98,7 @@ cache = [
 
 args = {
     "STACKABLE_USER_NAME": "stackable",
-    "STACKABLE_USER_UID": "1000",
-    "STACKABLE_USER_GID": "1000",
+    "STACKABLE_USER_UID": "782252253",
+    "STACKABLE_USER_GID": "574654813",
     "DELETE_CACHES": "true",
 }
--- a/conf.py
+++ b/conf.py
@@ -98,7 +98,7 @@ cache = [
 
 args = {
     "STACKABLE_USER_NAME": "stackable",
-    "STACKABLE_USER_UID": "782252253",
-    "STACKABLE_USER_GID": "574654813",
+    "STACKABLE_USER_UID": "782252253", # This is a random high id to not conflict with any existing user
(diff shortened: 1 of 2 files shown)


conf.py
```diff
@@ -98,7 +98,7 @@
 
 args = {
     "STACKABLE_USER_NAME": "stackable",
-    "STACKABLE_USER_UID": "1000",
-    "STACKABLE_USER_GID": "1000",
+    "STACKABLE_USER_UID": "782252253",
```
sbernauer, line 101: Can you please commented in code how you got these numbers?
Picked at random? If so please add a comment or link to best practices

lfrancke: I can add a comment, yes.
Just like the previous ones these are random, yes.

lfrancke: [`bf5a6a4` (#916)](https://github.com/stackabletech/docker-images/pull/916/commits/bf5a6a4cf3cd8ec53bb5281a15dd5cbcb4e80a6a)

sbernauer: Thanks!

sbernauer (approved):

lfrancke:
# Release Notes

The default user id and group id used by all our docker images was 1000. This has now been changed to high random numbers (782252253 / 574654813 respectively). Additionally our operators also hardcoded these values. This also [changed](https://github.com/stackabletech/issues/issues/651) which means that the user ids and group ids will change as of this release. This is for security purposes to reduce the chance of overlaps with existing ids on the host.

PR #918, opened 2024-11-07
docs(airflow): Document the need for installed packages

sbernauer:
# Description

```[tasklist]
```

<details>

```shell
```

</details>

### DIFF
--- a/airflow/Dockerfile
+++ b/airflow/Dockerfile
@@ -99,29 +99,29 @@ ENV AIRFLOW_USER_HOME_DIR=/stackable
 ENV PATH=$PATH:/bin:$HOME/app/bin
 ENV AIRFLOW_HOME=$HOME/airflow
 
-# Update image and install python
-RUN <<EOF
-microdnf update
-microdnf install \
-  ca-certificates \
-  cyrus-sasl \
-  # Needed for the gitsync functionality
-  git \
-  libpq \
-  # Needed for authentication of clients against LDAP servers
-  openldap \
-  openldap-clients \
-  # We need the openssh libs for the gitsync functionality (the clone target could be e.g. git@github.com:org/repo.git)
-  openssh-clients \
-  openssl-libs \
-  openssl-pkcs11 \
-  # Airflow needs Python
-  python${PYTHON} \
-  socat \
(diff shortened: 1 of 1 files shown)


airflow/Dockerfile
```diff
@@ -43,9 +43,8 @@ RUN microdnf update && \
         # Needed to modify the SBOM
         jq && \
     microdnf clean all && \
-    rm -rf /var/cache/yum
-
-RUN <<EOF
+    rm -rf /var/cache/yum && \
+bash <<EOF
```
razvan, line 47: Why are we now using both heredocs and individual RUN statements?

sbernauer: Sorry, I didn't fully get that. The only part that I see changed is now re-using the `bash` instead of spawning 2 separate bash invocations. The result (one RUN instead of two) should save disk space in the runners.

lfrancke: I'm against that as mentioned last week as it just introduces another concept. Just do what the other operators do even if it's not as pretty otherwise we have more inconsistencies.

razvan: > Sorry, I didn't fully get that. The only part that I see changed is now re-using the `bash` instead of spawning 2 separate bash invocations. The result (one RUN instead of two) should save disk space in the runners.

Yes, that one RUN command can also be simplified by putting the entire script in a single heredoc string like in this [example](https://github.com/stackabletech/docker-images/blob/812fdcf06c901654ba1facb95ef1a53d9deec0e6/hbase/Dockerfile#L150) 

It saves space *and*  is easier to read.

razvan: in other words, why is the `microdnf` call outside the <<EOF string ?

sbernauer: Reverted the optimization again in 296d1b905d8acea690bfedd0420b5fedb0525558, so that this Pr only focuses on adding some comments

sbernauer:
Build passed locally

lfrancke:
I don't want to have this style. Not because I don't like it but because I value consistency more.
I'll put up a follow-up PR to "fix" this (I know that it's subjective).

PR #929, opened 2024-11-11
fix(hive): CVE-2023-34455

razvan:
# Description

```[tasklist]
```

<details>

```shell
```

</details>

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -66,6 +66,7 @@ All notable changes to this project will be documented in this file.
 - spark: Fix CVE-2024-36114 in Spark 3.5.1 by upgrading a dependency.
   Spark 3.5.2 is not affected. ([#921])
 - trino: Correctly report Trino version ([#881]).
+- hive: Fix CVE-2023-34455 in Hive 4.0.0 by excluding snappy-java from the build ([#929])
 
 [#783]: https://github.com/stackabletech/docker-images/pull/783
 [#797]: https://github.com/stackabletech/docker-images/pull/797
@@ -111,6 +112,7 @@ All notable changes to this project will be documented in this file.
 [#919]: https://github.com/stackabletech/docker-images/pull/919
 [#920]: https://github.com/stackabletech/docker-images/pull/920
 [#921]: https://github.com/stackabletech/docker-images/pull/921
+[#929]: https://github.com/stackabletech/docker-images/pull/929
 
 ## [24.7.0] - 2024-07-24
 
--- a/hive/stackable/patches/4.0.0/04-exclude-snappy.patch
+++ b/hive/stackable/patches/4.0.0/04-exclude-snappy.patch
@@ -1,0 +1,27 @@
+Fix CVE-2023-34455
+see https://github.com/stackabletech/vulnerabilities/issues/558
+and https://github.com/stackabletech/vulnerabilities/issues/862
+
+Exclude snappy-java 1.1.8 from the standalone-metastore artifact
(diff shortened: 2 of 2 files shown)


NickLarsenNZ (dismissed):
LGTM

razvan:
Merged main and moved  changelog entries up to the **Unreleased** section.

lfrancke:
Instead of relying on getting a newer dependency "by accident" we'd need to actually properly depend on the new version.

razvan:
> Instead of relying on getting a newer dependency "by accident" we'd need to actually properly depend on the new version.

A newer version is already  configured in the parent [pom.xml](https://github.com/apache/hive/blob/183f8cb41d3dbed961ffd27999876468ff06690c/pom.xml#L204).

Here we exclude the vulnerable version that is coming with `hadoop-common`.

lfrancke:
But if we have a newer version in the parent we don't need to exclude it here, do we?
I'm sorry for the late response :(

razvan:
> But if we have a newer version in the parent we don't need to exclude it here, do we?

It's been a while so I had to start from the beginning.

Currently, without this patch,  the vulnerable snappy jar is **indeed** not part of the image (at list not as a stand-alone jar file). It is listed in the CycloneDX output, that's why the vulnerability is reported.

Proof:
```
$ docker run -ti --entrypoint bash --rm docker.stackable.tech/stackable/hive:4.0.0-stackable24.11.1
...
$ find /stackable/hadoop/ -name '*snappy*'
/stackable/hadoop/share/hadoop/common/lib/snappy-java-1.1.10.4.jar
/stackable/hadoop/share/hadoop/hdfs/lib/snappy-java-1.1.10.4.jar
$ find /stackable/hive-metastore/ -name '*snappy*'
$
$ grep snappy /stackable/hive-metastore/apache-hive-metastore-4.0.0.cdx.json
      "group" : "org.xerial.snappy",
      "name" : "snappy-java",
      "description" : "snappy-java: A fast compression/decompression library",
      "purl" : "pkg:maven/org.xerial.snappy/snappy-java@1.1.8.2?type=jar",
          "url" : "https://github.com/xerial/snappy-java"
          "url" : "https://github.com/xerial/snappy-java"
      "bom-ref" : "pkg:maven/org.xerial.snappy/snappy-java@1.1.8.2?type=jar"
        "pkg:maven/org.xerial.snappy/snappy-java@1.1.8.2?type=jar"
        "pkg:maven/org.xerial.snappy/snappy-java@1.1.8.2?type=jar",
      "ref" : "pkg:maven/org.xerial.snappy/snappy-java@1.1.8.2?type=jar",
```
This patch removes it from the CycloneDX file too. The confusion arises from the fact that the Cyclone report is generated from the original Hive pom.xml but we actually copy the Hadoop binaries that we build ourselves.

Now I think that a better approach would be to classify the vulnerability as a _false positive_ in SecObserve and close this PR.

WDYT?

lfrancke:
Okay. Understood. In that case I'm fine with either option. I just wanted to understand.

We can merge this PR (maybe with an extended comment on why we exclude it) or close it in SecObserve.

Please continue as you see fit.
@dervoeti FYI

dervoeti:
> Now I think that a better approach would be to classify the vulnerability as a _false positive_ in SecObserve and close this PR.

Sounds good to me.

I also checked the image and did not find the affected Snappy version anywhere, I even checked whether it's included in any other "fat" JAR, which was not the case. The vulnerability should go away completely once we generate our own version of Hadoop and depend on that in Hive's pom.xml, which will hopefully happen soon-ish.

PR #931, opened 2024-11-12
Spark now depends on HBase and should be rebuilt automatically as well

lfrancke:
# Description

Spark now depends on HBase and should be rebuilt automatically as well.


NickLarsenNZ (approved):
LGTM

PR #952, opened 2024-12-11
feat (tools): Add the package util-linux-core

### LINKED ISSUE #796: Add the package util-linux to the base image
Sometimes it is necessary to override the command of a stacklet with a custom script, or it is convenient to use the Stackable image for a custom Job. This is easier if more "standard" tools are installed, e.g. the `util-linux` package.

In a concrete case, a customer wanted a side-car container in the OPA pod which forwards the decision logs to a SIEM tool via syslog. A proper solution would be to configure the Vector aggregator accordingly, but on the one hand, the extra step with the aggregator was not desired, and on the other hand, Vector does not yet support syslog sinks ([vectordev/vector#6863](https://www.github.com/vectordotdev/vector/issues/6863)). So the implemented solution was to use the `logger` command to forward these logs. Unfortunately, this command is not available in the OPA image because the `util-linux` package is not installed. Therefore, a custom image had to be built.

Suggested change:

```shell
RUN microdnf install \
    # util-linux contains the command logger which allows to enter messages into the system log.
    # cpe:2.3:a:kernel:util-linux:*:*:*:*:*:*:*:*
    # https://nvd.nist.gov/vuln/search/results?isCpeNameSearch=true&query=cpe%3A2.3%3Aa%3Akernel%3Autil-linux%3A*%3A*%3A*%3A*%3A*%3A*%3A*%3A
    util-linux
```

siegfriedweber:
# Description

Add the package util-linux-core to th tools image.

util-linux-core contains a basic set of Linux utilities, including the command logger which allows to enter messages into the system log.

Closes #796 

```[tasklist]
```

<details>

```shell
```

</details>

### DIFF
--- a/tools/Dockerfile
+++ b/tools/Dockerfile
@@ -26,6 +26,12 @@ RUN microdnf update && \
     iputils \
     openssl \
     tar \
+    # util-linux-core contains a basic set of Linux utilities, including
+    # the command logger which allows to enter messages into the system
+    # log.
+    # cpe:2.3:a:kernel:util-linux:*:*:*:*:*:*:*:*
+    # https://nvd.nist.gov/vuln/search/results?isCpeNameSearch=true&query=cpe%3A2.3%3Aa%3Akernel%3Autil-linux%3A*%3A*%3A*%3A*%3A*%3A*%3A*%3A
+    util-linux-core \
     zip && \
     microdnf clean all && \
     rm -rf /var/cache/yum
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,9 @@ All notable changes to this project will be documented in this file.
 
 - nifi: Add NiFi hadoop Azure and GCP libraries ([#943]).
 - base: Add containerdebug tool ([#928]).
+- tools: Add the package util-linux-core ([#952]).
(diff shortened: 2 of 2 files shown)


tools/Dockerfile
```diff
@@ -26,6 +26,12 @@ RUN microdnf update && \
     iputils \
     openssl \
     tar \
+    # util-linux-core contains a basic set of Linux utilities, including
+    # the command logger which allows to enter messages into the system
+    # log.
+    # cpe:2.3:a:kernel:util-linux:*:*:*:*:*:*:*:*
+    # https://nvd.nist.gov/vuln/search/results?isCpeNameSearch=true&query=cpe%3A2.3%3Aa%3Akernel%3Autil-linux%3A*%3A*%3A*%3A*%3A*%3A*%3A*%3A
```
siegfriedweber, line 33: I added the link to `util-linux` instead of `util-linux-core` because `util-linux-core` seems not to be an existing product in the NIST NVD.

lfrancke (dismissed):

lfrancke:
Unfortunately it has conflicts now :(

lfrancke (approved):

PR #960, opened 2024-12-17
Change style of Airflow to be consistent with other Dockerfiles

lfrancke:
# Description

Follow-up to https://github.com/stackabletech/docker-images/pull/918


sbernauer (approved):
Thanks!

PR #985, opened 2025-01-29
revert(script): Change JMX Exporter source upload back to Jar upload

NickLarsenNZ:
Reverts the `upload_new_jmx_exporter_version.sh` script from #687 (it was missed in #933).

[Slack discussion](https://stackable-workspace.slack.com/archives/C02FZ581UCD/p1738149918664689)

PR #986, opened 2025-01-29
fix(script): Download jmx_exporter jar and sha256 sum file from github

NickLarsenNZ:
Correction from https://github.com/stackabletech/docker-images/pull/985.

The old location doesn't have the new Jar.
The script now also checks the sha256 sum before uploading.


siegfriedweber (approved):
LGTM

Issue #971, opened 2025-01-13
chore(druid): Update container image to 31.0.1 ahead of Stackable Release 25.3.0 #


sbernauer:
Part of https://github.com/stackabletech/docker-images/issues/964

- Add: `468` (or `469` if that is out)

> [!TIP]
> Please add the `scheduled-for/2025-03` label, and add to the [Stackable Engineering][1] project.
>
> [1]: https://github.com/orgs/stackabletech/projects/10

```[tasklist]
### Update tasks (trino)
- [ ] Update `versions.py` to reflect the agreed upon versions in the spreadsheet (including the removal of old versions).
- [ ] Upload new version (see `trino/*.sh` scripts).
- [ ] Update `versions.py` to the latest supported version of JVM (base and devel).
- [ ] Update other dependencies if applicable (eg: jmx_exporter, opa_authorizer, storage_connector, etc).
- [ ] Check other operators (getting_started / kuttl) for usage of the versions. Add the PR(s) to the list below.
- [ ] Update the version in demos. Add the PR(s) to the list below.
```

```[tasklist]
### Update tasks (trino-cli)
- [ ] Update `trino-cli/versions.py` to reflect the agreed upon versions in the spreadsheet (including the removal of old versions).
- [ ] Upload new version (see `trino-cli/upload_new_trino_version.sh` scripts).
- [ ] Update `versions.py` to the latest supported version of JVM (base and devel).
- [ ] Check other operators (getting_started / kuttl) for usage of the versions. Add the PR(s) to the list below.
- [ ] Update the version in demos. Add the PR(s) to the list below.
```

```[tasklist]
### Related Pull Requests
- [ ] _Link to the docker-images PR (product update)_
- [ ] _Link to the operator PR (getting_started / kuttl)_
- [ ] _Link to any other operator PRs (getting_started / kuttl)_
- [ ] _Link to demo PR (raise against the `next` branch)_
- [ ] _Link to the Release Notes PR in the documentation repo (if not a comment below)_
```

> [!TIP]
> Delete any items that do not apply so that all applicable items can be checked.
> For example, if you add release notes to the documentation repository, you do not need the latter two criteria.

This list should be completed by the assignee(s), once respective PRs have been merged. Once all items have been checked, the issue can be moved into _Development: Done_.

```[tasklist]
### Acceptance
- [ ] Can build image (either locally, or in CI)
- [ ] Kuttl smoke tests passes (either locally, or in CI)
[...]

PR #996, opened 2025-02-07
fix(trino): Restore trino-storage patches

NickLarsenNZ:
# Description

Part of https://github.com/stackabletech/docker-images/issues/971.

- Restore trino patches that were mistakenly removed in #993.
- Split trino-storage-connector into separate top level build directory.
  - This makes it easier to patch with patchable.
- Patch trino-storage-connector 469.


trino-storage-connector/upload_new_trino_storage_version.sh
Techassi, line 1: question: Should we rename this file to `upload_new_trino_storage_connector_version.sh`?

trino/Dockerfile
```diff
@@ -1,45 +1,11 @@
...
-cd trino-storage-${STORAGE_CONNECTOR}-src
-./patches/apply_patches.sh ${STORAGE_CONNECTOR}
-
-# Upstream builds are marked as -SNAPSHOT, even for release builds
-mvn versions:set -DnewVersion=${STORAGE_CONNECTOR}
-
-# We need to use ./mvnw instead of mvn to get a recent maven version (which is required to build Trino)
-./mvnw --batch-mode --no-transfer-progress package -DskipTests -Dmaven.gitcommitid.skip=true
-EOF
+FROM stackable/image/trino-storage-connector AS trino-storage-connector-image
```
Techassi, line 4: note: I think we should add a comment here explaining what this does.

Techassi: question: Also, do we want to stick with `trino-storage-connector-image` or use `trino-storage-connector` instead for consistency?

NickLarsenNZ: > note: I think we should add a comment here explaining what this does.

I can, but it is the same across every image - so maybe it needs documenting higher up? Like in the README?

> question: Also, do we want to stick with trino-storage-connector-image or use trino-storage-connector instead for consistency?

I think we stick with `trino-storage-connector-image` which is consistent (with others that are on the right side of the `AS`.

But also, we can just remove `FROM` and call the `stackable/image/trino-storage-connector` context directly from `COPY`.

NickLarsenNZ: We will defer this to a future PR, as we would like to:
- Restructure the README to be more useful.
- Cleanup the mixture of `_` and `-` separators used on bake-contexts and `ARG`s.
- Document how the bake contexts work (and how they relate to `ARG`s).

Techassi (approved):

Issue #1011, opened 2025-02-28
Organize dependency build processes


dervoeti:
Implementation of https://github.com/stackabletech/decisions/issues/44

See sub-issues for the list of goals.

Non-Goal:
 - Publishing artifacts (e.g., JARs in a Maven repository) is out of scope for this issue. This will be addressed when implementing [stackabletech/issues#674](https://github.com/stackabletech/issues/issues/674).

Issue #677, opened 2024-05-15
Unify patch system (in Dockerfiles)


nightkr:
Currently, each product has its own copy of our overall patch system. It would be nice to unify them into one implementation in one location that all of them can reference.

lfrancke:
I'm not sure how much we want this issue to spiral out of control but we can also use this to discuss how we want to handle patches and build-from-source in the first place.

We can also fork the upstream repos and build the things including patches there once, push them to a registry and only fetch them from there during image build....

dervoeti:
Implemented by https://github.com/stackabletech/docker-images/pull/1032

The apply_patches.sh scripts are now replaced by `patchable checkout`, which checks out the source repository (defined in the `patchable.toml` files) and applies our custom patches on top of it.

lfrancke:
⚱ RIP `apply_patches.sh` we hardly knew ye!

PR #1032, opened 2025-03-18
feat: move patch apply logic to patchable

dervoeti:
# Description

This is a draft PR to solve https://github.com/stackabletech/docker-images/issues/677. I replaced the `apply_patches.sh` script with Patchable for Zookeeper and Druid.

We decided in https://github.com/stackabletech/decisions/issues/44 that all sources should be mirrored on Github (to be implemented by https://github.com/stackabletech/docker-images/issues/1011). 

My idea was that we can just reuse the existing `checkout` logic, since it does exactly what we want: checkout a repo at a specific commit and apply a bunch of .patch files on top of it. In this PR, the original Apache repositories are used, those would need to be replaced by our mirror repos.

I added an optional flag `--images_repo_root` to manually specify a custom root for the `images` repo instead of relying on the auto-detection that Patchable currently does. We could also just use the auto-detection, but then we'd probably need to run `git init && touch .patchable` before running the `patchable checkout` command inside the container, which is a bit awkward. The extra flag isn't super pretty either though and I couldn't get around using `PathBuf` in the implementation, hence I made this a draft PR to serve as a PoC. This is just one way we could replace the patch scripts with Patchable. Maybe this can be solved in a more elegant way, I didn't find one yet though.

```[tasklist]
```

<details>

```shell
```

</details>

### DIFF
--- a/druid/Dockerfile
+++ b/druid/Dockerfile
@@ -23,10 +23,8 @@ microdnf update
 #   This requirement is documented in docs/development/build.md and version 5.1 or later is required.
 #   UBI 9 ships with 5.4.x so that should be fine
 #
-# patch: Required for the apply-patches.sh script
 microdnf install \
-python-pyyaml \
-patch
+python-pyyaml
 
 microdnf clean all
 rm -rf /var/cache/yum
@@ -35,8 +33,7 @@ EOF
 USER ${STACKABLE_USER_UID}
 WORKDIR /stackable
 
-COPY --chown=stackable:0 druid/stackable/patches/apply_patches.sh /stackable/apache-druid-${PRODUCT}-src/patches/apply_patches.sh
-COPY --chown=stackable:0 druid/stackable/patches/${PRODUCT} /stackable/apache-druid-${PRODUCT}-src/patches/${PRODUCT}
+COPY --chown=stackable:0 druid/stackable/patches/${PRODUCT} /stackable/apache-druid-${PRODUCT}-src/druid/stackable/patches/${PRODUCT}
 
 # Cache mounts are owned by root by default
 # We need to explicitly give the uid to use which is hardcoded to "1000" in stackable-base
@@ -50,10 +47,8 @@ RUN --mount=type=cache,id=maven-${PRODUCT},uid=${STACKABLE_USER_UID},target=/sta
     --mount=type=cache,id=npm-${PRODUCT},uid=${STACKABLE_USER_UID},target=/stackable/.npm \
     --mount=type=cache,id=cache-${PRODUCT},uid=${STACKABLE_USER_UID},target=/stackable/.cache \
     <<EOF
-curl "https://repo.stackable.tech/repository/packages/druid/apache-druid-${PRODUCT}-src.tar.gz" | tar -xzC .
 cd apache-druid-${PRODUCT}-src
-./patches/apply_patches.sh ${PRODUCT}
-
+cd "$(/stackable/patchable --images-repo-root=. checkout druid ${PRODUCT})" && \
 mvn --batch-mode --no-transfer-progress clean install -Pdist,stackable-bundle-contrib-exts -DskipTests -Dmaven.javadoc.skip=true
 mv distribution/target/apache-druid-${PRODUCT}-bin/apache-druid-${PRODUCT} /stackable/
 mv distribution/target/bom.json /stackable/apache-druid-${PRODUCT}/apache-druid-${PRODUCT}.cdx.json
--- a/rust/patchable/src/main.rs
+++ b/rust/patchable/src/main.rs
@@ -107,6 +107,10 @@ impl ProductVersionContext<'_> {
 struct Opts {
     #[clap(subcommand)]
     cmd: Cmd,
+
+    /// Specify a custom root directory for the images repository
+    #[clap(long)]
+    images_repo_root: Option<PathBuf>,
 }
 
 #[derive(clap::Parser)]
@@ -165,7 +169,7 @@ enum Cmd {
         pv: ProductVersion,
     },
 
-    /// Shwos the images repository root
+    /// Shows the images repository root
     ImagesDir,
 }
 
@@ -201,6 +205,8 @@ pub enum Error {
     FindImagesRepo { source: repo::Error },
     #[snafu(display("images repository has no work directory"))]
     NoImagesRepoWorkdir,
+    #[snafu(display("images repository root at {path:?} is not a directory"))]
+    ImagesRepoRootDirCheck { path: PathBuf },
 
     #[snafu(display("failed to fetch patch series' base commit"))]
     FetchBaseCommit { source: repo::Error },
@@ -263,8 +269,19 @@ fn main() -> Result<()> {
     .context(ConfigureGitLoggingSnafu)?;
 
     let opts = <Opts as clap::Parser>::parse();
-    let images_repo = repo::discover_images_repo(".").context(FindImagesRepoSnafu)?;
-    let images_repo_root = images_repo.workdir().context(NoImagesRepoWorkdirSnafu)?;
+    let images_repo_root_pathbuf = match opts.images_repo_root {
+        Some(path) => {
+            if !path.is_dir() {
+                return ImagesRepoRootDirCheckSnafu { path }.fail();
+            }
+            path
+        }
+        None => {
+            let images_repo = repo::discover_images_repo(".").context(FindImagesRepoSnafu)?;
+            images_repo.workdir().context(NoImagesRepoWorkdirSnafu)?.to_owned()
+        }
+    };
+    let images_repo_root = images_repo_root_pathbuf.as_path();
     match opts.cmd {
         Cmd::Checkout { pv, base_only } => {
             let ctx = ProductVersionContext {
--- a/stackable-base/Dockerfile
+++ b/stackable-base/Dockerfile
@@ -59,6 +59,23 @@ rustup toolchain install
 cargo auditable --quiet build --release && cargo cyclonedx --all --spec-version 1.5 --describe binaries
 EOF
 
+FROM product-utils-builder AS patchable
+
+# Copy source code of patchable
+COPY rust/patchable/ /patchable/rust/patchable
+# Copy workspace files
+COPY Cargo.* /patchable/
+
+RUN <<EOF
+microdnf update --assumeyes
+microdnf install --assumeyes openssl-devel pkg-config
+cd /patchable
+. "$HOME/.cargo/env"
+rustup toolchain install
+cargo auditable --quiet build --release && cargo cyclonedx --all --spec-version 1.5 --describe binaries
+microdnf clean all
+EOF
+
 # Find the latest version at https://catalog.redhat.com/software/containers/ubi9/ubi-minimal/615bd9b4075b022acc111bf5?container-tabs=gti
 # IMPORTANT: Make sure to use the "Manifest List Digest" that references the images for multiple architectures
 # rather than just the "Image Digest" that references the image for the selected architecture.
@@ -203,6 +220,10 @@ COPY --from=config-utils --chown=${STACKABLE_USER_UID}:0 /config-utils/config-ut
 # **containerdebug**
 # Debug tool that logs generic system information.
 COPY --from=containerdebug --chown=${STACKABLE_USER_UID}:0 /containerdebug/target/release/containerdebug /stackable/containerdebug
+
+# **patchable**
+# Tool for patch management
+COPY --from=patchable --chown=${STACKABLE_USER_UID}:0 /patchable/target/release/patchable /stackable/patchable
 
 ENV PATH="${PATH}:/stackable"
 
--- a/zookeeper/Dockerfile
+++ b/zookeeper/Dockerfile
@@ -17,10 +17,9 @@ USER ${STACKABLE_USER_UID}
 WORKDIR /stackable
 
 # Download ZooKeeper sources from our own repo
-RUN curl "https://repo.stackable.tech/repository/packages/zookeeper/apache-zookeeper-${PRODUCT}.tar.gz" | tar -xzC . && \
-    # Apply any required patches
-    patches/apply_patches.sh ${PRODUCT} && \
-    cd /stackable/apache-zookeeper-${PRODUCT}/ && \
+RUN mkdir -p zookeeper/stackable && mv patches zookeeper/stackable/ && \
+    BUILD_SRC_DIR="$(/stackable/patchable --images-repo-root=. checkout zookeeper ${PRODUCT})" && \
+    cd "${BUILD_SRC_DIR}" && \
     # Exclude the `zookeeper-client` submodule, this is not needed and has c parts
     # that created all kinds of issues for the build container
     mvn --batch-mode --no-transfer-progress -pl "!zookeeper-client/zookeeper-client-c" clean install checkstyle:check spotbugs:check -DskipTests -Pfull-build && \
@@ -29,7 +28,7 @@ RUN curl "https://repo.stackable.tech/repository/packages/zookeeper/apache-zooke
     # Unpack the archive which contains the build artifacts from above. Remove some
     # unused files to shrink the final image size.
     tar xvzf /stackable/apache-zookeeper-${PRODUCT}-bin.tar.gz && \
-    mv /stackable/apache-zookeeper-${PRODUCT}/zookeeper-assembly/target/bom.json /stackable/apache-zookeeper-${PRODUCT}-bin/apache-zookeeper-${PRODUCT}.cdx.json && \
+    mv "${BUILD_SRC_DIR}/zookeeper-assembly/target/bom.json" /stackable/apache-zookeeper-${PRODUCT}-bin/apache-zookeeper-${PRODUCT}.cdx.json && \
     rm -rf /stackable/apache-zookeeper-${PRODUCT}-bin/docs && \
     rm /stackable/apache-zookeeper-${PRODUCT}-bin/README_packaging.md && \
     # Download the JMX exporter jar from our own repo


rust/patchable/src/main.rs
```diff
@@ -263,8 +269,19 @@ fn main() -> Result<()> {
     .context(ConfigureGitLoggingSnafu)?;
 
     let opts = <Opts as clap::Parser>::parse();
-    let images_repo = repo::discover_images_repo(".").context(FindImagesRepoSnafu)?;
-    let images_repo_root = images_repo.workdir().context(NoImagesRepoWorkdirSnafu)?;
+    let images_repo_root_pathbuf = match opts.images_repo_root {
```
nightkr, line 272: nit:
```suggestion
    let images_repo_root = match opts.images_repo_root {
```
We can always borrow it where we need to.

rust/patchable/src/main.rs
```diff
@@ -165,7 +169,7 @@ enum Cmd {
         pv: ProductVersion,
     },
 
-    /// Shwos the images repository root
+    /// Shows the images repository root
```
nightkr, line 172: Oops

rust/patchable/src/main.rs
```diff
@@ -263,8 +269,19 @@ fn main() -> Result<()> {
     .context(ConfigureGitLoggingSnafu)?;
 
     let opts = <Opts as clap::Parser>::parse();
-    let images_repo = repo::discover_images_repo(".").context(FindImagesRepoSnafu)?;
-    let images_repo_root = images_repo.workdir().context(NoImagesRepoWorkdirSnafu)?;
+    let images_repo_root_pathbuf = match opts.images_repo_root {
+        Some(path) => {
+            if !path.is_dir() {
+                return ImagesRepoRootDirCheckSnafu { path }.fail();
+            }
```
nightkr, line 276: Not sure we actually need this validation - any actual file access will fail anyway (and I'd say that, for something like `init`, this is a perfectly fine place to be in).

rust/patchable/src/main.rs
nightkr, line 272: nit: the `match` here is fine, but this would also be a nice place to simplify with [`Option::map_or_else`](https://doc.rust-lang.org/stable/std/option/enum.Option.html#handling-a-result-based-fallback)

stackable-base/Dockerfile
```diff
@@ -204,6 +221,10 @@ COPY --from=config-utils --chown=${STACKABLE_USER_UID}:0 /config-utils/config-ut
 # Debug tool that logs generic system information.
 COPY --from=containerdebug --chown=${STACKABLE_USER_UID}:0 /containerdebug/target/release/containerdebug /stackable/containerdebug
 
+# **patchable**
+# Tool for patch management
+COPY --from=patchable --chown=${STACKABLE_USER_UID}:0 /patchable/target/release/patchable /stackable/patchable
```
nightkr, line 226: We probably don't want to include patchable in the product images we ship to users... Arguably we'd ideally have a separate `stackable-devel` image... but that then gets into a funky diamond dependency for `java-devel`.. :/

rust/patchable/src/main.rs
```diff
@@ -107,6 +107,10 @@ impl ProductVersionContext<'_> {
 struct Opts {
     #[clap(subcommand)]
     cmd: Cmd,
+
+    /// Specify a custom root directory for the images repository
+    #[clap(long)]
+    images_repo_root: Option<PathBuf>,
```
nightkr, line 113: Hm.. do we want to set the root here, or the individual paths? Then again, the enforced convention seems to slot in pretty well for our Dockerfiles, even if they need to be adapted to it.

druid/Dockerfile
```diff
@@ -35,8 +33,7 @@ EOF
 USER ${STACKABLE_USER_UID}
 WORKDIR /stackable
 
-COPY --chown=stackable:0 druid/stackable/patches/apply_patches.sh /stackable/apache-druid-${PRODUCT}-src/patches/apply_patches.sh
-COPY --chown=stackable:0 druid/stackable/patches/${PRODUCT} /stackable/apache-druid-${PRODUCT}-src/patches/${PRODUCT}
+COPY --chown=stackable:0 druid/stackable/patches/${PRODUCT} /stackable/apache-druid-${PRODUCT}-src/druid/stackable/patches/${PRODUCT}
```
nightkr, line 36:
```suggestion
COPY --chown=stackable:0 druid/stackable/patches/${PRODUCT} /stackable/src/druid/stackable/patches/${PRODUCT}
```

The `blah-1.2.3-src` format tends to imply that it's an extract of the upstream source archive (which is what we'd put at `$ROOT/blah/worktree/1.2.3`, not at `$ROOT`).

druid/Dockerfile
```diff
@@ -50,10 +47,8 @@ RUN --mount=type=cache,id=maven-${PRODUCT},uid=${STACKABLE_USER_UID},target=/sta
     --mount=type=cache,id=npm-${PRODUCT},uid=${STACKABLE_USER_UID},target=/stackable/.npm \
     --mount=type=cache,id=cache-${PRODUCT},uid=${STACKABLE_USER_UID},target=/stackable/.cache \
     <<EOF
-curl "https://repo.stackable.tech/repository/packages/druid/apache-druid-${PRODUCT}-src.tar.gz" | tar -xzC .
 cd apache-druid-${PRODUCT}-src
-./patches/apply_patches.sh ${PRODUCT}
-
+cd "$(/stackable/patchable --images-repo-root=. checkout druid ${PRODUCT})" && \
```
nightkr, line 51: This could then be

```bash
cd "$(/stackable/patchable --images-repo-root=src checkout druid ${PRODUCT})" && \
```

zookeeper/Dockerfile
```diff
@@ -17,10 +17,9 @@ USER ${STACKABLE_USER_UID}
 WORKDIR /stackable
 
 # Download ZooKeeper sources from our own repo
-RUN curl "https://repo.stackable.tech/repository/packages/zookeeper/apache-zookeeper-${PRODUCT}.tar.gz" | tar -xzC . && \
-    # Apply any required patches
-    patches/apply_patches.sh ${PRODUCT} && \
-    cd /stackable/apache-zookeeper-${PRODUCT}/ && \
+RUN mkdir -p zookeeper/stackable && mv patches zookeeper/stackable/ && \
```
nightkr, line 20: I assume this is an earlier revision? I think I prefer how it's done in druid

rust/patchable/src/main.rs
dervoeti: Yeah I think keeping the convention here as well is nicer. Also made the refactoring of the `patchable` source code easier I think. No strong opinion though.

rust/patchable/src/main.rs
nightkr: With moe it becomes:

```rs
    let images_repo_root = opts.images_repo_root.map_or_else(
        || {
            let images_repo = repo::discover_images_repo(".").context(FindImagesRepoSnafu)?;
            images_repo
                .workdir()
                .context(NoImagesRepoWorkdirSnafu)
                .map(Path::to_owned)
        },
        Ok,
    )?;
```

I wouldn't call it a significant regression for error handling, but I'm also fine with keeping it as is.

druid/Dockerfile
dervoeti: I changed the name of the source directory, but we have to `cd` into it anyway (or adapt all the following `mvn` and `mv` commands as well, but I think `cd` is easier)

nightkr: That'd be covered by cding into the worktree right after anyway, no? (That's the actual dir it needs to run in, anyway)

rust/patchable/src/main.rs
dervoeti: Ah, yeah it's not as complicated as I imagined it to be, but at least to me the `match` variant is more readable

druid/Dockerfile
dervoeti: Oh! You're right. I'll change it.

rust/patchable/src/main.rs
nightkr: In fact, it would probably be cleaner to just move it into the PVC without borrowing:

```patch
diff --git a/rust/patchable/src/main.rs b/rust/patchable/src/main.rs
index d32f9b4b..c1adcb53 100644
--- a/rust/patchable/src/main.rs
+++ b/rust/patchable/src/main.rs
@@ -5,11 +5,7 @@ mod repo;
 mod utils;
 
 use core::str;
-use std::{
-    fs::File,
-    io::Write,
-    path::{Path, PathBuf},
-};
+use std::{fs::File, io::Write, path::PathBuf};
 
 use git2::{Oid, Repository};
 use serde::{Deserialize, Serialize};
@@ -35,12 +31,12 @@ struct ProductVersionConfig {
     base: Oid,
 }
 
-struct ProductVersionContext<'a> {
+struct ProductVersionContext {
     pv: ProductVersion,
-    images_repo_root: &'a Path,
+    images_repo_root: PathBuf,
 }
 
-impl ProductVersionContext<'_> {
+impl ProductVersionContext {
     fn load_config(&self) -> Result<ProductVersionConfig> {
         let path = &self.config_path();
         tracing::info!(
@@ -271,10 +267,12 @@ fn main() -> Result<()> {
         Some(path) => path,
         None => {
             let images_repo = repo::discover_images_repo(".").context(FindImagesRepoSnafu)?;
-            images_repo.workdir().context(NoImagesRepoWorkdirSnafu)?.to_owned()
+            images_repo
+                .workdir()
+                .context(NoImagesRepoWorkdirSnafu)?
+                .to_owned()
         }
     };
-    let images_repo_root = images_repo_root.as_path();
     match opts.cmd {
         Cmd::Checkout { pv, base_only } => {
             let ctx = ProductVersionContext {
```

stackable-base/Dockerfile
dervoeti: Hmm. For now, it would be enough to put patchable in `java-devel`, since I think we're currently only patching products that are built using that image. But once we start patching other products like Superset or OPA, this won't be sufficient.
Maybe I'll try to create a `stackable-devel` image that can used in the build stage by all products. Just a small layer on top of `stackable-base`. `java-devel` can then probably just start with `FROM stackable/image/stackable-devel`. I'll play around with it tomorrow.

nightkr: We could also just copy in the binaries from stackable-devel. But java-devel should be fine for now too.

dervoeti: I created `stackable-devel` now, works fine so far :slightly_smiling_face:

nightkr: :+1:

rust/patchable/src/main.rs
dervoeti: Yep, is changed now

dervoeti:
@nightkr It's ready for another round of feedback 🙂 Once it's in a good shape I'll start adapting the patch processes of the other products.

nightkr (commented):
Largely LGTM

dervoeti:
@nightkr I migrated all Dockerfiles now and tested all builds using the latest versions of the product, worked fine.

druid/Dockerfile
```diff
@@ -44,10 +43,12 @@ COPY --chown=${STACKABLE_USER_UID}:0 druid/stackable/patches/${PRODUCT} /stackab
 # are still working in the cache directory.
 
 RUN --mount=type=cache,id=maven-${PRODUCT},uid=${STACKABLE_USER_UID},target=/stackable/.m2/repository \
-    --mount=type=cache,id=npm-${PRODUCT},uid=${STACKABLE_USER_UID},target=/stackable/.npm \
-    --mount=type=cache,id=cache-${PRODUCT},uid=${STACKABLE_USER_UID},target=/stackable/.cache \
-    <<EOF
+  --mount=type=cache,id=npm-${PRODUCT},uid=${STACKABLE_USER_UID},target=/stackable/.npm \
+  --mount=type=cache,id=cache-${PRODUCT},uid=${STACKABLE_USER_UID},target=/stackable/.cache \
+  <<EOF
+curl "https://repo.stackable.tech/repository/packages/druid/apache-druid-${PRODUCT}-src.tar.gz" | tar -xzC .
```
nightkr, line 49: This is moot when patchable is responsible for downloading the sources, it still builds fine if I remove this line.

hbase/Dockerfile
```diff
@@ -44,21 +44,18 @@ RUN --mount=type=cache,id=maven-hbase-${PRODUCT},uid=${STACKABLE_USER_UID},targe
...
 # The release scripts of HBase also run the build twice (three times in fact, once again to build the site which we skip here).
 # I chose to replicate that exact behavior for consistency so please don't merge the two mvn runs into one unless you really know what you're doing!
 mvn --batch-mode --no-transfer-progress -Dhadoop.profile=3.0 -Dhadoop-three.version=${HADOOP} clean install -DskipTests
 mvn --batch-mode --no-transfer-progress -Dhadoop.profile=3.0 -Dhadoop-three.version=${HADOOP} install assembly:single -DskipTests -Dcheckstyle.skip=true -Prelease
 tar -xzf hbase-assembly/target/hbase-${PRODUCT}-bin.tar.gz -C /stackable/
 mv hbase-assembly/target/bom.json /stackable/hbase-${PRODUCT}/hbase-${PRODUCT}.cdx.json
-rm -rf /stackable/hbase-${PRODUCT}-src
+
+# Remove sources
+(cd .. && rm -r ${PRODUCT})
```
nightkr, line 57: Is this because we run out of space? I see that we already did this before, but it would've been nice to have a motivation in the comment.

hbase/Dockerfile
```diff
@@ -155,30 +151,32 @@ WORKDIR /stackable
...
-chmod +x patches/apply_patches.sh
-patches/apply_patches.sh ${HBASE_OPERATOR_TOOLS} hbase-operator-tools-${HBASE_OPERATOR_TOOLS}-src
+cd "$(/stackable/patchable --images-repo-root=src checkout hbase-operator-tools ${HBASE_OPERATOR_TOOLS})"
+cd hbase-hbck2
+git init
+git config user.email "fake.commiter@stackable.tech"
+git config user.name "Fake commiter"
+git commit --allow-empty --message "Fake commit, so that we can create a tag"
+git tag ${HBASE_OPERATOR_TOOLS}
+cd ..
```
nightkr, line 161: Hm, is this a new requirement somewhere?

hive/Dockerfile
```diff
@@ -22,46 +22,51 @@ ARG STACKABLE_USER_UID
 # This can be used to speed up builds when disk space is of no concern.
 ARG DELETE_CACHES="true"
 
+# Copy patches into the builder
+COPY --chown=${STACKABLE_USER_UID}:0 hive/stackable/patches/${PRODUCT} /stackable/src/hive/stackable/patches/${PRODUCT}
+# Copy JMX config into the builder
+COPY --chown=${STACKABLE_USER_UID}:0 hive/stackable/jmx /stackable/jmx
+# Copy the start script into the builder
+COPY --chown=${STACKABLE_USER_UID}:0 hive/stackable/bin /stackable/bin
```
nightkr, line 30: If we don't use it during the build then it's nice if we can copy it in afterwards, so we don't have to recompile it if they are modified.

superset/Dockerfile
```diff
@@ -115,10 +114,8 @@ RUN python3 -m venv /stackable/app \
     pip install --no-cache-dir /tmp/opa_authorizer-0.1.0-py3-none-any.whl
 
 COPY --from=statsd_exporter-builder /statsd_exporter/statsd_exporter /stackable/statsd_exporter
-COPY superset/stackable/patches /patches
 
 RUN <<EOF
-/patches/apply_patches.sh ${PRODUCT}
```
nightkr, line 121: Hm, half-inclined to say "just let patchable manage this too, for consistency", even if there are no patches *currently*

trino/Dockerfile
```diff
@@ -9,39 +9,25 @@ ARG PRODUCT
...
-./patches/apply_patches.sh ${PRODUCT}
+cd "$(/stackable/patchable --images-repo-root=src checkout trino ${PRODUCT})"
 
 # Trino is using something (git-commit-id-plugin in the past, maybe something else now) that is
 # reading the Git history and searches for a tag to pull the version from. It sounds weird to me
 # why someone would do that over just picking the version from the pom.xml, but they propably
 # have their reasons. See e.g. https://github.com/trinodb/trino/discussions/18963.
 # So we fake it till we make it and create a Git repo and the correct tag. The trino-operator
 # smoke test checks that "select version()" is working.
+rm -rf .git
```
nightkr, line 30: Do we need to blow away the git repo here? Wouldn't it just be enough to tag the leaf commit that patchable created?

hbase/Dockerfile
dervoeti: Hmm we could also just remove the Git repo here, I think I'll do that. It looks like a Maven plugin is trying to get the commit id, which fails after checking out the source with patchable:
```
Failed to execute goal pl.project13.maven:git-commit-id-plugin:2.2.5:revision (get-the-git-infos) on project hbase-hbck2: Could not complete Mojo execution... Error: Could not get HEAD Ref, are you sure you have some commits in the dotGitDirectory?
```

trino/Dockerfile
dervoeti: It should be sufficient, but the Maven plugin has problems with Git worktrees, it fails with:
```
Failed to execute goal io.github.git-commit-id:git-commit-id-maven-plugin:9.0.1:revision (default) on project trino-testing-services: Could not get HEAD Ref, are you sure you have some commits in the dotGitDirectory (currently set to /stackable/src/trino/patchable-work/product-repo/worktrees/470)?
```
Basically the same problem as with hbase-operator-tools. There's an [issue](https://github.com/git-commit-id/git-commit-id-maven-plugin/issues/215) for this, people mention a workaround that uses native Git, but it did not work for me (neither for Trino nor for hbase-operator-tools).

druid/Dockerfile
dervoeti: Good catch! Fixed.

hbase/Dockerfile
dervoeti: Yeah I think this is done to reduce the layer size. We could clarify it in all our builds, but this will change soon, since we want to include the sourcecode used to build the product in the image (last task of [this issue](https://github.com/stackabletech/docker-images/issues/1011)), Lars mentioned this to me again last week that we should do this. So I'd say we leave it like this for now.

superset/Dockerfile
dervoeti: We don't have patches for our Python based products at the moment, I'm not even sure how the proper way to apply these patches would look like, so I thought "let's cross that bridge when we get to it".

trino/Dockerfile
nightkr: Ahh, that's funky. And it seems like setting `nativegit` runs git in the wrong folder - also not very helpful.

I'd probably update the comment then to mention the worktree issue, and remove the `-rf` from the `rm`? ..we might also want to think about updating their jgit (https://github.com/git-commit-id/git-commit-id-maven-plugin/issues/215#issuecomment-2397662054) somehow, since this is still an issue for local builds.

superset/Dockerfile
nightkr: Aah, didn't realize that we install it straight from PyPI. Yeah, fair to call it out of scope for this PR then.

hbase/Dockerfile
dervoeti: This is changed now so that the `.git` file is removed, similar to the Trino build.

trino/Dockerfile
dervoeti: Removed the `-rf` now and added a comment.

hive/Dockerfile
dervoeti: Yep, good point, moved it to the `final` stage since both files are not needed at build-time.

nightkr (approved):
LGTM if it builds for you.

Issue #961, opened 2024-12-18
chown/chmod massively balloons the size of our docker images


nightkr:
This was a regression introduced in #849 (and its followups, tracked in https://github.com/stackabletech/issues/issues/645).

When you `chmod`/`chown` a file owned by a prior layer, its contents are copied into the new layer, even though they are identical. Since clients download the whole layer stack, they get to download both versions.

As an experiment, removing the chmod/chowns from the nifi image (added in #890):

```patch
diff --git a/trino/Dockerfile b/trino/Dockerfile
index 68383bdb..9f6f2f42 100644
--- a/trino/Dockerfile
+++ b/trino/Dockerfile
@@ -154,8 +154,8 @@ ln -s /stackable/jmx/jmx_prometheus_javaagent-${JMX_EXPORTER}.jar /stackable/jmx
 
 # All files and folders owned by root group to support running as arbitrary users.
 # This is best practice as all container users will belong to the root group (0).
-chown -R ${STACKABLE_USER_UID}:0 /stackable
-chmod -R g=u /stackable
+# chown -R ${STACKABLE_USER_UID}:0 /stackable
+# chmod -R g=u /stackable
 EOF
 
 # ----------------------------------------
```

took the total image size from 5.4GiB to 1.6GiB (according to dive).

That aside, chowning all of /stackable is a tad overly aggressive. The products shouldn't need access to overwrite their own source code, for example.

siegfriedweber:
The files in `/stackable` are often copied from a builder image with `COPY --chown=${STACKABLE_USER_UID}:0 --from=builder ...`, so the `chown -R ${STACKABLE_USER_UID}:0 /stackable` isn't actually the problem, the `chmod -R g=u /stackable` is. This can be solved by adding the parameter `--chmod` to the `COPY` statement:

```Dockerfile
COPY \
    --chown=${STACKABLE_USER_UID}:0 \
    --chmod=u=g \
    --from=builder \
    ...
```

In my test, the layer size was reduced from 1.2 GB to 11 MB without changing the `RUN` statement. If `/stackable/containerdebug` would also be copied with this parameter, then the layer size would be negligible.

maltesander:
I did some investigation and testing / fixing PRs. Most important ones are NiFi and Trino.
- [ ] Airflow(2.10.4)   - 1.94 GB ~ ok?
- [ ] Druid (31.0.1)    - 1.85 GB ~ ok?
[...]

PR #1053, opened 2025-04-07
Use check permissions script and consolidation

maltesander:
# Description

part of https://github.com/stackabletech/docker-images/issues/961

No image size reduction, just consolidation.

- Does remove the recursive chmod/chown in the final image and attempts non recursive chmod/chown or doing as much as possible in the builder step.
- Uses check-permissions-ownership script to validate any wrong permissions / ownership in the /stackable folder


siegfriedweber (dismissed):
LGTM

siegfriedweber (approved):

PR #1057, opened 2025-04-14
ci: Add reusable image build workflow

Techassi:
> [!NOTE]
> Airflow test run succeeded: https://github.com/stackabletech/docker-images/actions/runs/14442557563

This adds a reusable image build workflow which encapsulates all required steps to build a (product) image into a single file. This workflow can then be used by each product / tool which massively reduces duplicated code.

This PR also removes the no longer supported `build-cache-password` input from the `build-product-image` action (caused failing builds due to cache corruption).


NickLarsenNZ (approved):
LGTM

PR #1065, opened 2025-04-24
fix: add missing patchable config for Kafka 3.8.0

dervoeti:
Builds are failing since https://github.com/stackabletech/docker-images/pull/1032 because the Patchable config for Kafka 3.8.0 is missing


NickLarsenNZ (approved):
LGTM

Issue #1049, opened 2025-04-03
Migrate all mirrored sources from Nexus to GitHub repositories


dervoeti:
Update the “upload new version” process, which currently relies on shell scripts, to use GitHub instead. Ensure all historical versions are preserved if necessary.

dervoeti:
Results from a quick discussion with @nightkr:
- I'll create a draft implemenation for `patchable init --sync-mirror`. Forks need to be created manually.
- Do we want to put the forks in a separate GH org?
- Do we want to use gh forks or create our own mirror repos?

dervoeti:
We changed some things regarding `patchable` in #1070:
 - A default mirror repository can be specified for each product (`patchable init product druid --upstream https://github.com/apache/druid.git --default-mirror https://github.com/stackabletech/druid.git`)
 - The source code of each product version can be mirrored in that repository (`patchable init version druid 31.0.1 --base=druid-31.0.1 --mirror`)
 - `patchable` will go ahead and copy the branch or tag specified in `--base` to the mirror repository as part of the `init version` subcommand
 - `patchable checkout` (used e.g. inside Dockerfiles) will clone from that mirror repository and apply the patches on top of it

When adding a new product that should be mirrored, the mirror repository needs to be created manually on Github (using the "fork" functionality).

PR #1070, opened 2025-04-30
feat: patchable init mirroring

dervoeti:
Fixes https://github.com/stackabletech/docker-images/issues/1049

This adds the `--mirrored` flag to `patchable init`. It assumes that a mirror repository already exist, pushes the ref specified via `--base` to the mirror and configures the mirrorer repo as `upstream` in patchable.toml.
As discussed with @lfrancke, we want to fork the repos in the `stackabletech` org and use forks instead of clones (the latter decision was not a strong opinion, I decided to go with forks since they would facilitate back-contributions).

I created mirror repositories (forks in our case) for all our products and synced the tags for all versions we use. I updated all patchable.toml files to use the mirror repositories (with the exception of Omid `1.1.3-SNAPSHOT`, which is a temporary version we don't ship).

I had to drop the usage of `FETCH_HEAD`, since it caused problems when running `patchable init` for a tag/branch that for some reason was already present in the local Git repo. The problem is that `FETCH_HEAD` is empty in that case, I think that's due to no changes being fetched.


nightkr:
As a nomenclature thing, I think I prefer calling this mirroring instead.

dervoeti:
> As a nomenclature thing, I think I prefer calling this mirroring instead.

Had the same thought when opening the PR :slightly_smiling_face: Will refactor it.

rust/patchable/src/main.rs
```diff
@@ -414,8 +441,93 @@ fn main() -> Result<()> {
...
+                    &product_repo,
+                    &base,
+                    &upstream
+                )
                 .context(FetchBaseCommitSnafu)?;
+
+                tracing::info!(commit = %base_commit_oid, "fetched base commit OID");
+
+                // Add fork remote
+                let temp_fork_remote = "patchable_fork_push";
```
nightkr: You should be able to use [`remote_anonymous`](https://docs.rs/git2/latest/git2/struct.Repository.html#method.remote_anonymous) instead of creating and deleting a dummy remote.

rust/patchable/src/main.rs
```diff
@@ -414,8 +441,93 @@ fn main() -> Result<()> {
...
+                    }
+                });
+
+                let mut push_options = git2::PushOptions::new();
+                push_options.remote_callbacks(callbacks);
+
+                // Check if the reference is a tag or branch by inspecting the git repository
+                let refspec = {
+                    let tag_ref = format!("refs/tags/{}", base);
+                    let is_tag = product_repo
```
nightkr, line 499: I half think we should just decide on tags with a common format here.

rust/patchable/src/main.rs
```diff
@@ -414,8 +441,93 @@ fn main() -> Result<()> {
...
+                // Add progress tracking for push operation
+                let span_push = tracing::info_span!("pushing");
+                span_push.pb_set_style(&utils::progress_bar_style());
+                let _ = span_push.enter();
+                let mut quant_push = utils::Quantizer::percent();
+                callbacks.push_transfer_progress(move |current, total, _| {
+                    if total > 0 {
+                        quant_push.update_span_progress(current, total, &span_push);
+                    }
+                });
```
nightkr, line 491: We should probably break this stuff out into a utils module.

rust/patchable/src/main.rs
```diff
@@ -414,8 +441,93 @@ fn main() -> Result<()> {
...
+                    })?;
+
+                product_repo.remote_delete(temp_fork_remote)
+                    .context(DeleteRemoteSnafu { name: temp_fork_remote.to_string() })?;
+
+                tracing::info!("successfully pushed base ref to fork");
+
+                (base_commit_oid, fork_url)
+            } else {
+                (repo::resolve_and_fetch_commitish(&product_repo, &base, &upstream).context(FetchBaseCommitSnafu)?, upstream)
```
nightkr, line 528: The initial `resolve_and_fetch` is still the same, no? The new mirroring stuff is "just" an extra step afterwards.

rust/patchable/src/repo.rs
```diff
@@ -180,7 +180,9 @@ pub fn resolve_and_fetch_commitish(
                     &[commitish],
                     Some(
                         FetchOptions::new()
-                            .update_fetchhead(true)
```
nightkr, line 183: Curious about the fetchhead issues you ran into, do you have a repro for that? Does that happen with the current `main`, or only with (the rest of) this PR applied?

druid/stackable/patches/30.0.0/patchable.toml
```diff
@@ -1,2 +1,2 @@
-upstream = "https://github.com/apache/druid.git"
+upstream = "https://github.com/stackabletech/druid.git"
```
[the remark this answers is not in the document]
dervoeti: Yep, good idea. Done.

rust/patchable/src/main.rs
dervoeti: Oh right, fixed!

rust/patchable/src/main.rs
dervoeti: Correct. Didn't notice this after my last refactoring :slightly_smiling_face: It's simplified now.

rust/patchable/src/main.rs
dervoeti: Did some refactoring [here](https://github.com/stackabletech/docker-images/pull/1070/commits/3034966ebf4feb1e465a375804bfd22fb8e7960b)

rust/patchable/src/main.rs
dervoeti: Hmm but we can't really know if the upstream repository uses tags in a specific format (or if it uses tags at all). Zookeeper uses `release-3.9.2`, Trino `470`, trino-storage uses `v470`, ...
Or do you mean we should re-tag the commit in our mirror?

rust/patchable/src/repo.rs
dervoeti: Interestingly it works with the current `main` branch. As soon as the line `.download_tags(git2::AutotagOption::Auto)` is added, it starts breaking. Steps to reproduce:

Add the `.download_tags(git2::AutotagOption::Auto)` line after `.update_fetchhead`, then run
```
cargo patchable init zookeeper 3.9.2 --upstream=https://github.com/apache/zookeeper.git --base=release-3.9.2
rm zookeeper/stackable/patches/3.9.2/patchable.toml
cargo patchable init zookeeper 3.9.2 --upstream=https://github.com/apache/zookeeper.git --base=release-3.9.2
```
For me, the second command fails with:
```
Error: failed to fetch patch series' base commit

Caused by these errors (recent errors listed first):
  1: failed to find commit "release-3.9.2" in "/home/voeti/docker-images/zookeeper/patchable-work/product-repo/"
  2: corrupted loose reference file: FETCH_HEAD; class=Reference (4)
```

I'm not sure why FETCH_HEAD is empty. I'm also not sure why/if `download_tags` is really required to keep the tag around in `patchable-work/product-repo/refs/tags`. I tried a few things (disabled pruning) but I couldn't make it work, so I decided to remove the `update_fetchhead` and just use `commitish` instead of `FETCH_HEAD`, which circumvented the problem.

rust/patchable/src/main.rs
nightkr: Mostly that we should always use tags, instead of checking whether upstream uses a tag or a branch.

dervoeti: The mirror-functionality implemented in this PR currently takes the upstream reference (tag or branch) and copies it to the mirror repo. That would break if we limit it to tags only and the upstream repo wouldn't have any tags (just branches). I think all our current upstream repos use tags though, so we could do it. Should we rename `--base` to `--tag` then?

nightkr: I don't see how branches would change anything for us - we request a commitish (could be a branch, tag, commit id, whatever) and get a commit id back (as well as the backing objects required to actually use it). We create a tag named after the commitish (or the version id? that's the separate discussion) pinned to the commit id that we got.

That said: I don't think branch-as-base is a valid use case that we ought to cater to, if there isn't a stable tag then you should probably confirm the commit hash yourself.

nightkr: Actually, I guess one use case I _could_ see for branch-as-base would be for tracking upstream development branches - but I think that's less "we should cater to this" and more "we shouldn't break it".

dervoeti: Ah, now I see. So we always create a tag in our mirror repo, no matter what the upstream base ref is and just name that tag like the commitish (which will usually be a tag anyway). Sounds good, I'll implement that.

dervoeti: @nightkr Done in https://github.com/stackabletech/docker-images/pull/1070/commits/42e9cb47761a80e116957e1f95ae59502179006d
That also allowed me to revert the change regarding `FETCH_HEAD` / `download_tags`.
Did some tests with upstream tags and branches, worked fine.

rust/patchable/src/repo.rs
nightkr: No longer relevant: https://github.com/stackabletech/docker-images/pull/1070#discussion_r2073413726

rust/patchable/src/main.rs
```diff
@@ -414,13 +437,71 @@ fn main() -> Result<()> {
...
+            let base_commit = repo::resolve_and_fetch_commitish(&product_repo, &base, &upstream).context(FetchBaseCommitSnafu)?;
+            let mut upstream_mirror = None;
+
+            if mirrored {
+                // Parse e.g. "https://github.com/apache/druid.git" into "druid"
+                let repo_name = upstream.split('/').last().map(|repo| repo.trim_end_matches(".git")).context(ParseUpstreamUrlSnafu { url: &upstream })?;
+
+                ensure!(!repo_name.is_empty(), ParseUpstreamUrlSnafu { url: &upstream });
+
+                let mirror_url = format!("https://github.com/stackabletech/{}.git", repo_name);
```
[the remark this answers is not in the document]
nightkr: We might still want to let versions override the fields in the product config, though..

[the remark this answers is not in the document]
nightkr: Another alternative: product-scope `patchable.toml` is _only_ used by `patchable init`, relevant fields are copied into the the version's `patchable.toml` by `patchable init` (always `upstream`, `mirror` only when ran with `--mirror`).

rust/patchable/src/main.rs
```diff
@@ -414,13 +437,71 @@ fn main() -> Result<()> {
...
+                // Push the base commit to the mirror
+                tracing::info!(commit = %base_commit, base = base, url = mirror_url, "pushing commit to mirror");
+                let mut callbacks = git2::RemoteCallbacks::new();
+                callbacks.credentials(|_url, username_from_url, _allowed_types| {
+                    git2::Cred::credential_helper(
+                        &git2::Config::open_default().unwrap(), // Use default git config
+                        _url,
+                        username_from_url,
+                    )
+                });
```
nightkr, line 466: Do we actually need this? Feels like a lot of faff just to get the default behaviour...

rust/patchable/src/main.rs
```diff
@@ -414,13 +437,71 @@ fn main() -> Result<()> {
...
-            let base_commit = repo::resolve_and_fetch_commitish(&product_repo, &base, &upstream)
-                .context(FetchBaseCommitSnafu)?;
+            let base_commit = repo::resolve_and_fetch_commitish(&product_repo, &base, &upstream).context(FetchBaseCommitSnafu)?;
+            let mut upstream_mirror = None;
+
+            if mirrored {
+                // Parse e.g. "https://github.com/apache/druid.git" into "druid"
+                let repo_name = upstream.split('/').last().map(|repo| repo.trim_end_matches(".git")).context(ParseUpstreamUrlSnafu { url: &upstream })?;
+
+                ensure!(!repo_name.is_empty(), ParseUpstreamUrlSnafu { url: &upstream });
```
nightkr, line 447: Can use `Option::filter` instead of `ensure!` if we want to use the same `ParseUpstreamUrlSnafu` error for both cases.

rust/patchable/src/utils.rs
```diff
@@ -94,3 +87,14 @@ pub mod oid_serde {
...
     }
 }
+
+/// Sets up progress tracking for Git operations with a progress bar.
+pub fn setup_progress_tracking(span: tracing::Span) -> (tracing::Span, Quantizer) {
+    span.pb_set_style(&ProgressStyle::with_template(
+        "{span_child_prefix}{spinner} {span_name}{{{span_fields}}} {wide_msg} {bar:40} {percent:>3}%",
+    )
+    .expect("hard-coded template should be valid"));
+    let _ = span.enter();
```
nightkr, line 97: This will be dropped when `setup_progress_tracking` returns, you'd need to return the guard. You'd also need to use [`Span::entered`](https://docs.rs/tracing/latest/tracing/struct.Span.html#method.entered) instead to avoid borrow issues.

But that said, it also feels like a pretty heavy thing for a "setup" function to do; I think I'd prefer for the caller to be responsible for entering the span.

rust/patchable/src/main.rs
```diff
@@ -197,6 +203,18 @@ pub enum Error {
         path: PathBuf,
     },
 
+    #[snafu(display("failed to parse upstream URL {url:?} to extract repository name"))]
+    ParseUpstreamUrl { url: String },
```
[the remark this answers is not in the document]
dervoeti: Hmm. I wanted to avoid adding `regex` just for this. Should we provide a valid example in the error message? Like: "Expected a URL in the form of https://github.com/org/reponame"?
I know validation could be stronger here but it felt like overkill.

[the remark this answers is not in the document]
nightkr: I think the example is good, but ideally https://github.com/stackabletech/docker-images/pull/1070#discussion_r2073586179 would let us get rid of the parsing entirely.

rust/patchable/src/utils.rs
dervoeti: Fixed, callers are responsible for entering the span now.

rust/patchable/src/main.rs
dervoeti: Ah yeah, done.

rust/patchable/src/main.rs
nightkr: Fair :/ Maybe worth a comment?

rust/patchable/src/main.rs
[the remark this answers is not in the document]
dervoeti: I now implemented your first suggestion (`upstream`/`mirror` are separated from `base`).
> We might still want to let versions override the fields in the product config

I implemented this at first in [this commit](https://github.com/stackabletech/docker-images/pull/1070/commits/694e5c758db2c6f817243b0454c1b3ca997d616f), but it bloated the code and I felt like we don't really need this, so I threw it out.

I still need to update the existing patchable.toml files, will do so once we think this implementation is fine.

rust/patchable/src/main.rs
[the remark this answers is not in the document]
dervoeti: > We'd probably need some kind of URL rewrite, since we don't want to force people to pull through SSH, or to only allow pushing over SSH.

Hmm. Maybe just store the base identifier of the repository in the configuration file, like `github.com/user/repository.git`.
Push/Pull via HTTPS by default, but add a new `--ssh` flag. By default, the base identifier would be transformed into `https://github.com/user/repository.git`. With `--ssh` set, it would be transformed into `git@github.com:user/repository.git`. WDYT?
We could even make the SSH username configurable, but that's probably not needed since we currently rely on Github exclusively. We can always extend it later.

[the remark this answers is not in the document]
nightkr: Hm, yeah.. It doesn't even need to be _that_ complicated, `git@github.com:user/repository.git` is equivalent to `ssh://git@github.com/user/repository.git`, so I guess it's just a prefix we need to add. Annoyingly, [`Url::set_schema`] doesn't allow changing the schema between `ssh` and `https`, so we're still stuck in string manipulation land.

>  We could even make the SSH username configurable, but that's probably not needed since we currently rely on Github exclusively. We can always extend it later.

Yeah, agreed.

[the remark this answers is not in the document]
nightkr: (Though if I am to be _really_ annoying, one argument for defaulting to storing the full HTTPS URL would be that you could do local testing by setting a local path like `file:blah`.)

rust/patchable/src/main.rs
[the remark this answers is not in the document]
nightkr: Hm, I think mirroring is the kicker there.. I'd still want `init` to require `--mirror` before actually mirroring anything, it feels weird and dangerous that we suddenly start calling out to external services _by default_ for an otherwise local tool. And that requires the version config to at least note whether or not that version is mirrored.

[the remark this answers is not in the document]
dervoeti: Well it only does so by default if `mirror` was explicitly specified in the product config. If we enable mirroring only via the explicit `--mirror` flag, users might forget to set the flag, in which case that version won't be mirrored. And I think it makes sense to configure mirroring at the product-level. I don't really see a use case for mirroring only some versions. Especially for us, it's an internal tool and I think we want to mirror all versions we create patches for.
I see your point regarding the usage of external services, that might be surprising. Maybe we can highlight this more in the README. Or maybe require an explicit confirmation before pushing to the mirror?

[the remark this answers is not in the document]
nightkr: We could flag it as a warning, or enforce it in CA.

> I don't really see a use case for mirroring only some versions. Especially for us, it's an internal tool and I think we want to mirror all versions we create patches for.

For things merged to `main`, absolutely. But that doesn't cover everything; I think it's premature to demand mirroring when *starting* the work of bringing up support for a new version. Especially for externals, people experimenting with nightly builds for whatever reason, etc.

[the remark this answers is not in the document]
dervoeti: Hmm ok, good points. So I could move the `mirror` property from product- to version-level. Product config would just contain `upstream` and version config would contain `base` and optionally `mirror`? And then I'd restore the `--mirror` flag.
That way we still would get around implementing the merge logic to support overrides by the version-level config (it worked but is probably not worth the overhead).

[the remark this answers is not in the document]
nightkr: I'd do something like

```patch
diff --git a/rust/patchable/src/main.rs b/rust/patchable/src/main.rs
index c5b972ee..4c6925a1 100644
--- a/rust/patchable/src/main.rs
+++ b/rust/patchable/src/main.rs
@@ -27,11 +27,13 @@ struct ProductVersion {
 #[derive(Deserialize, Serialize)]
 struct ProductConfig {
     upstream: String,
-    mirror: Option<String>,
+    target_mirror: Option<String>,
 }
 
 #[derive(Deserialize, Serialize)]
 struct ProductVersionConfig {
+    mirror: Option<String>,
+
     #[serde(with = "utils::oid_serde")]
     base: Oid,
 }
@@ -163,6 +165,9 @@ enum Cmd {
         /// Refs (such as tags and branches) will be resolved to commit IDs.
         #[clap(long)]
         base: String,
+
+        #[clap(long)]
+        mirror: bool,
     },
 
     /// Shows the patch directory for a given product version
@@ -211,6 +216,10 @@ pub enum Error {
         path: PathBuf,
     },
 
+    #[snafu(display(
+        "mirroring requested, but target-mirror is not configured in product configuration"
+    ))]
+    MirrorNotConfigured,
     #[snafu(display("failed to add temporary mirror remote for {url:?}"))]
     AddMirrorRemote { source: git2::Error, url: String },
     #[snafu(display("failed to push commit {commit} (as {refspec}) to mirror {url:?}"))]
@@ -312,10 +321,10 @@ fn main() -> Result<()> {
             let base_commit = repo::resolve_and_fetch_commitish(
                 &product_repo,
                 &version_config.base.to_string(),
-                product_config
-                    .mirror
-                    .as_deref()
-                    .unwrap_or(&product_config.upstream),
+                version_config.mirror.as_deref().unwrap_or_else(|| {
+                    tracing::warn!("this product version is not mirrored, re-init it with --mirror before merging it");
+                    &product_config.upstream
+                }),
             )
             .context(FetchBaseCommitSnafu)?;
             let base_branch = ctx.base_branch();
@@ -425,7 +434,7 @@ fn main() -> Result<()> {
             );
         }
 
-        Cmd::Init { pv, base } => {
+        Cmd::Init { pv, base, mirror } => {
             let ctx = ProductVersionContext {
                 pv,
                 images_repo_root,
@@ -449,7 +458,9 @@ fn main() -> Result<()> {
                 .context(FetchBaseCommitSnafu)?;
             tracing::info!(?base, base.commit = ?base_commit, "resolved base commit");
 
-            if let Some(mirror_url) = config.mirror {
+            let mirror_url = if mirror {
+                let mirror_url = config.target_mirror.context(MirrorNotConfiguredSnafu)?;
+
                 // Add mirror remote
                 let mut mirror_remote =
                     product_repo
@@ -491,16 +502,25 @@ fn main() -> Result<()> {
                 mirror_remote
                     .push(&[&refspec], Some(&mut push_options))
                     .context(PushToMirrorSnafu {
-                        url: mirror_url,
+                        url: &mirror_url,
                         refspec: &refspec,
                         commit: base_commit,
                     })?;
 
                 tracing::info!("successfully pushed base ref to mirror");
+                Some(mirror_url)
+            } else {
+                tracing::warn!(
+                    "this version is not mirrored, re-run with --mirror before merging into main"
+                );
+                None
             };
 
             tracing::info!("saving version-level configuration");
-            let config = ProductVersionConfig { base: base_commit };
+            let config = ProductVersionConfig {
+                base: base_commit,
+                mirror: mirror_url,
+            };
             let config_path = ctx.version_config_path();
             if let Some(config_dir) = config_path.parent() {
                 std::fs::create_dir_all(config_dir)
diff --git a/rust/patchable/src/repo.rs b/rust/patchable/src/repo.rs
index e4aa500e..00602310 100644
--- a/rust/patchable/src/repo.rs
+++ b/rust/patchable/src/repo.rs
@@ -149,8 +149,10 @@ pub fn resolve_and_fetch_commitish(
                 "base commit not found locally, fetching from upstream"
             );
 
-            let (span_recv, mut quant_recv) = setup_progress_tracking(tracing::info_span!("receiving"));
-            let (span_index, mut quant_index) = setup_progress_tracking(tracing::info_span!("indexing"));
+            let (span_recv, mut quant_recv) =
+                setup_progress_tracking(tracing::info_span!("receiving"));
+            let (span_index, mut quant_index) =
+                setup_progress_tracking(tracing::info_span!("indexing"));
 
             let _ = span_recv.enter();
             let _ = span_index.enter();
```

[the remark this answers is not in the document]
dervoeti: Hmm ok looks good. Feel free to commit this to this branch :slightly_smiling_face:

Two ideas I had when looking over the changes:
 - Rename `target_mirror` to `default_mirror`?
 or:
 - Make the `mirror` inside `ProductVersionConfig` a bool (like `use_mirror`)?

I'm not saying these are good ideas, just wanted to mention them if you think some of this would improve it. I'm fine with either variant, feel free to just ignore them.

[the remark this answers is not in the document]
nightkr: 51b26eda86c6f43a420b2030cf94f0ba5442e3ee

rust/patchable/src/main.rs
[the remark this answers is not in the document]
dervoeti: I added SSH support in https://github.com/stackabletech/docker-images/pull/1070/commits/dd32e8aa5588d1ce6e258c4a6917ffac59d536cd
Authentication is only supported through SSH agent, it seems like the `git2` crate doesn't support `~/.ssh/config` (it uses libssh, not OpenSSH, https://github.com/rust-lang/git2-rs/pull/1031 might help). We could maybe also parse the SSH config ourselves with [ssh2-config](https://crates.io/crates/ssh2-config) but... oof.

rust/patchable/README.md
```diff
@@ -31,18 +31,30 @@ For more details, run `cargo patchable --help`.
...
+2. A version-level config file at `docker-images/<PRODUCT>/stackable/patches/<VERSION>/patchable.toml`
+
+The product-level config contains:
 - `upstream` - the URL of the upstream repository (such as `https://github.com/apache/druid.git`)
-- `base` - the commit hash of the upstream base commit (such as `7cffb81a8e124d5f218f9af16ad685acf5e9c67c`)
+- `mirror` - optional URL of a mirror repository (such as `https://github.com/stackabletech/druid.git`)
 
-### Template
+The version-level config contains:
+- `base` - the commit hash of the upstream base commit
```
nightkr, line 44: Oh right, should probably update this..

rust/patchable/src/main.rs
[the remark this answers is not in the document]
nightkr: Closing since we discuss SSH in https://github.com/stackabletech/docker-images/pull/1070#discussion_r2073569962.

rust/patchable/src/main.rs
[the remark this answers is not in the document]
nightkr: Yeah, that's fair.. Sucks that libgit2 doesn't seem to have a "just do whatever the git CLI does" option..

rust/patchable/README.md
dervoeti: Done in https://github.com/stackabletech/docker-images/pull/1070/commits/b101cd4b52ed9ef9856cd479981cce2b00c8f641

rust/patchable/src/utils.rs
```diff
@@ -1,8 +1,36 @@
...
+    let parsed_url = Url::parse(original_url).context(ParseUrlSnafu {
+        url: original_url,
+    })?;
+
+    let host = parsed_url.host_str().context(NoHostInUrlSnafu {
+        url: original_url,
+    })?;
+    let path = parsed_url.path().trim_start_matches('/');
+
+    Ok(format!("git@{}:{}", host, path))
```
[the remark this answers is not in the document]
dervoeti: Changed with https://github.com/stackabletech/docker-images/pull/1070/commits/9923c9dc84e252e9ef653941c6716af0cff4177a

nightkr (approved):

dervoeti:
@nightkr I would now go ahead and update all the `patchable.toml` files (create product-level configs and adapt the version-level configs). And update this PR to be compatible with the current `main` branch.

nightkr:
Sounds good to me.

dervoeti:
@nightkr This is done now. Product-level configs are all there and I removed `upstream` from version-level configs and added `mirror` instead.

nightkr:
All versions seem to check out fine (`ls **/stackable/patches | insert version { ls $in.name -s | filter { $in.type == "dir" } | get name } | insert product { $in.name | path dirname -n 2 } | flatten | each { cargo patchable checkout $in.product $in.version }`).

nightkr (approved):

dervoeti:
@nightkr I noticed on more thing :slightly_frowning_face: Some builds would fail, because we use `patchable checkout` inside Docker to checkout the source before building the product. For this, the product config is needed.
At least for some images we only copy the specific version directory, which of course does not include the product config. I'll change that, so that the whole `patches` directory is copied into the image, not just the directory for the specific version.

nightkr:
Ahh, right. Good catch.

dervoeti:
@nightkr Fixed in https://github.com/stackabletech/docker-images/pull/1070/commits/61486086d383ebeaf4f15f34290fbf36c2cde0c7

dervoeti:
@nightkr Reverted and fixed by copying only the product config

nightkr (approved):

PR #1124, opened 2025-05-21
chore: stop building kcat image

dervoeti:
# Description

We have an image `sdp/kcat` and also `sdp/kafka-testing-tools`, which seems to contain kcat (and nothing else really).
`sdp/kafka-testing-tools` is used in a kuttl test for kafka-operator.
As far as I see, the published `sdp/kcat` image is not used anywhere. The image neither has our usual labels nor an entrypoint.

I don't see a reason why we publish the kcat image or why anyone would use it (`kafka-testing-tools` seems to be the proper image that contains kcat since it has labels and an entrypoint), it looks like it's just an intermediate image that was published for some reason.

The plan is: Keep all the current `sdp/kcat` images around, just to be safe, but stop building new ones for upcoming releases, mainly to avoid confusion because we currently have two images that essentially contain the same thing (`kcat` and `kafka-testing-tools`). So this PR just removes the Github Action that publishes `sdp/kcat`.
This should be mentioned in the release notes.

Also: Fixed a bug in `.scripts/update_readme_badges.sh` that broke the pre-commit checks in this case (number of columns at the end was hardcoded).

> [!NOTE]

<details>

```shell
```

</details>


razvan (commented):
there is more

```
$ rg kcat --hidden
...
.github/workflows/build_kafka.yaml
16:      - kcat/**

.github/workflows/build_kafka-testing-tools.yaml
16:      - kcat/**
```

razvan (approved):
lgtm!

Issue #1069, opened 2025-04-29
Add source code to our images


dervoeti:
Distribute the exact source code (including custom patches!) that was used to build the product in the final images. Also include source code for non-products components if possible (kcat, Phoenix, nifi-opa-authorizer, ...).

lfrancke:
This is something that we should document in my opinion. I don't know where to be honest though. Somewhere in the vicinity of the SBOM docs maybe?
And I'd like to have release notes as well please.

I think this is fantastic and we should write about it.

dervoeti:
Regarding documentation:
Yeah it's worth mentioning, but I'm also not sure where. I could add a sentence or two here:
https://docs.stackable.tech/home/stable/guides/viewing-and-verifying-sboms/
Or maybe here:
https://docs.stackable.tech/home/stable/product-information/

Created a [draft PR](https://github.com/stackabletech/documentation/pull/743) for this.

### Release note
We've added source code snapshots directly into our container images to further improve transparency and the debugging experience. Each container image now includes `.tar.gz` archives containing the exact source code used to build that specific version of the product. They can be found within the `/stackable` directory of each container image.

lfrancke:
Good point on the operators. I think we should.

PR #1126, opened 2025-05-23
feat: include source code

dervoeti:
# Description

Fixes https://github.com/stackabletech/docker-images/issues/1069

This PR adds .tar.gz files containing the source code (including our custom patches) that was used to build the products into the container images. This should help with debugging and further increase transparency, while the costs are relatively low.

I also migrated a few components to fetch the source code using patchable. The goal is to use the same process everywhere to fetch source code (clone it from a Github repository that is part of the stackabletech org), even if we don't patch the source code.

The following components were migrated to use `patchable`:
 - protobuf: I had add the invocation of `./autogen.sh` (along with its dependencies, `automake` and `libtool`), since the code of the Github repo differed from the snapshot we have in Nexus (for example, in Nexus the `./configure` script is already present). This is also the exact [same way Hadoop does it](https://github.com/apache/hadoop/blob/949292eac6d26ddf5713bf2783d09c3317ffe695/dev-support/docker/pkg-resolver/install-protobuf.sh#L49-L50).
 - hdfs-utils
 - hbase-opa-authorizer
 - hbase/hbase-operator-tools: Just a slight change, I moved it inside the image from `/stackable/src/hbase-operator-tools` to `/stackable/src/hbase/hbase-operator-tools` to align with the structure in docker-images and all the other components
 - kcat
 - nifi-iceberg-bundle
 - NiFi opa-authorizer
 - OPA

> [!NOTE]

<details>

```shell
```

</details>


dervoeti:
@maltesander Thanks, I moved the `chmod` statements now

maltesander (approved):
Thanks, LGTM!

PR #1138, opened 2025-06-02
chore: Update the Airflow entrypoint script and remove the UID & GID check

lfrancke:
# Description

We copy the entrypoint script fromt he "official" Dockerfiles and we haven't updated that in two years.
This does just that and in addition it does disable the UID & GID check which we believe to be wrong. The GID doesn't need to be zero on Vanilla K8s and OpenShift handles it itself.

See https://github.com/stackabletech/issues/issues/651 for details


adwk67 (approved):
LGTM. I got a couple of OOM failures with the tests but that is not related to this PR.

PR #1159, opened 2025-06-11
feat: HBase resolvable endpoints

adwk67:
# Description

fixes: https://github.com/stackabletech/hbase-operator/issues/641

> [!WARNING]
> Merge this at the same time as https://github.com/stackabletech/hbase-operator/pull/639, otherwise tests will break!

> [!NOTE]

<details>

```shell
```

</details>


hbase/stackable/patches/2.6.1/0005-Allow-overriding-ipc-bind-port-and-use-alternative-p.patch
```diff
@@ -0,0 +1,195 @@
...
++  }
++
+   private void registerConfigurationObservers() {
+     configurationManager.registerObserver(this.rpcServices);
+     configurationManager.registerObserver(this);
+@@ -586,8 +592,7 @@ public class HMaster extends HRegionServer implements MasterServices {
+       registerConfigurationObservers();
+       Threads.setDaemonThreadRunning(new Thread(() -> TraceUtil.trace(() -> {
+         try {
+-          int infoPort = putUpJettyServer();
```
nightkr, line 40: Don't we still need to launch the jetty server? Maybe this is done somewhere else so it still works, but it feels like something risky to get rid of entirely in a patch.

```suggestion
-          putUpJettyServer();
```

hbase/stackable/patches/2.6.1/0005-Allow-overriding-ipc-bind-port-and-use-alternative-p.patch
```diff
@@ -0,0 +1,195 @@
...
+ import static org.apache.hadoop.hbase.master.cleaner.HFileCleaner.CUSTOM_POOL_SIZE;
+ import static org.apache.hadoop.hbase.util.DNS.MASTER_HOSTNAME_KEY;
+ 
+@@ -559,6 +560,11 @@ public class HMaster extends HRegionServer implements MasterServices {
+     return conf.get(MASTER_HOSTNAME_KEY);
+   }
+ 
++  @Override
++  protected int getUseThisPortInstead(Configuration conf) {
++    return conf.getInt(MASTER_PORT, this.rpcServices.getSocketAddress().getPort());
```
nightkr, line 30: One caveat here is that previously you could set `hbase.master.port=0` to automatically allocate the port number, and the advertisement would automatically be updated to the correct port.

This patch still _allows_ that behaviour, but requires you to instead set `hbase.master.ipc.port=0` and _unset_ `hbase.master.port`. A workaround would to use something like:

```suggestion
+    int port = conf.getInt(MASTER_PORT, 0);
+    return port != 0 ? port : this.rpcServices.getSocketAddress().getPort();
```

hbase/stackable/patches/2.6.1/0005-Allow-overriding-ipc-bind-port-and-use-alternative-p.patch
```diff
@@ -0,0 +1,195 @@
...
++      zooKeeper = new ZKWatcher(conf, getProcessName() + ":" + useThisPortInstead, this,
+         canCreateBaseZNode());
+       // If no master in cluster, skip trying to track one or look for a cluster status.
+       if (!this.masterless) {
+@@ -776,6 +783,10 @@ public class HRegionServer extends Thread
+     }
+   }
+ 
++  protected int getUseThisPortInstead(Configuration conf) {
++    return conf.getInt(REGIONSERVER_PORT, this.rpcServices.isa.getPort());
```
nightkr, line 110: Same thing here.

hbase/stackable/patches/2.6.1/0005-Allow-overriding-ipc-bind-port-and-use-alternative-p.patch
```diff
@@ -0,0 +1,195 @@
...
++
+   /**
+    * @deprecated since 2.4.0 and will be removed in 4.0.0. Use
+    *             {@link HRegionServer#UNSAFE_RS_HOSTNAME_DISABLE_MASTER_REVERSEDNS_KEY} instead.
+@@ -669,6 +674,8 @@ public class HRegionServer extends Thread
+       this.namedQueueRecorder = NamedQueueRecorder.getInstance(this.conf);
+       rpcServices = createRpcServices();
+       useThisHostnameInstead = getUseThisHostnameInstead(conf);
++      useThisPortInstead = getUseThisPortInstead(conf);
++      useThisInfoPortInstead = conf.getInt("hbase.info.port" , this.infoServer != null ? this.infoServer.getPort() : -1);
```
nightkr, line 83: And here.

nightkr, line 83: Do we want to pull `hbase.info.port` into a constant for consistency? For the same reason, we might want to split it into master/regionserver variants?

Also, directionally.. for RPC ports we follow the existing RPC _hostname_ convention of "leave the existing key as the advertised port, introduce a new key for overriding the bound port". I agree that overriding the advertised port instead makes more sense _in isolation_ (and is consistent with kafka et al), but we should probably try to be consistent (both with the rest of our patch and with existing HBase code).

hbase/stackable/patches/2.6.1/0005-Allow-overriding-ipc-bind-port-and-use-alternative-p.patch
```diff
@@ -0,0 +1,195 @@
...
+diff --git a/hbase-server/src/main/java/org/apache/hadoop/hbase/regionserver/RSRpcServices.java b/hbase-server/src/main/java/org/apache/hadoop/hbase/regionserver/RSRpcServices.java
+index b77fcf338a..1f5c9dd21f 100644
+--- a/hbase-server/src/main/java/org/apache/hadoop/hbase/regionserver/RSRpcServices.java
++++ b/hbase-server/src/main/java/org/apache/hadoop/hbase/regionserver/RSRpcServices.java
+@@ -1270,14 +1270,14 @@ public class RSRpcServices implements HBaseRPCErrorHandler, AdminService.Blockin
+       int port = conf.getInt(HConstants.MASTER_PORT, HConstants.DEFAULT_MASTER_PORT);
+       // Creation of a HSA will force a resolve.
+       initialIsa = new InetSocketAddress(hostname, port);
+-      bindAddress = new InetSocketAddress(conf.get("hbase.master.ipc.address", hostname), port);
++      bindAddress = new InetSocketAddress(conf.get("hbase.master.ipc.address", hostname), conf.getInt("hbase.master.ipc.port", port));
```
nightkr, line 184: Same constant question here tbh.

hbase/stackable/patches/2.6.1/0005-Allow-overriding-ipc-bind-port-and-use-alternative-p.patch
```diff
@@ -0,0 +1,195 @@
...
+     } else {
+       String hostname = DNS.getHostname(conf, DNS.ServerType.REGIONSERVER);
+       int port = conf.getInt(HConstants.REGIONSERVER_PORT, HConstants.DEFAULT_REGIONSERVER_PORT);
+       // Creation of a HSA will force a resolve.
+       initialIsa = new InetSocketAddress(hostname, port);
+       bindAddress =
+-        new InetSocketAddress(conf.get("hbase.regionserver.ipc.address", hostname), port);
++        new InetSocketAddress(conf.get("hbase.regionserver.ipc.address", hostname), conf.getInt("hbase.regionserver.ipc.port", port));
```
nightkr, line 192: And here.

hbase/stackable/patches/2.6.1/0005-Allow-overriding-ipc-bind-port-and-use-alternative-p.patch
```diff
@@ -0,0 +1,195 @@
...
++
+   private void setupSignalHandlers() {
+     if (!SystemUtils.IS_OS_WINDOWS) {
+       HBasePlatformDependent.handle("HUP", (number, name) -> {
+@@ -958,7 +969,7 @@ public class HRegionServer extends Thread
+       }
+       // Setup RPC client for master communication
+       this.rpcClient = RpcClientFactory.createClient(conf, clusterId,
+-        new InetSocketAddress(this.rpcServices.isa.getAddress(), 0),
++        new InetSocketAddress(0),
```
nightkr, line 121: I don't think the old behaviour here made sense _in general_, but I assume someone needed it at some point. Do we want to have a flag for reintroducing it?

(IMO, this probably makes less sense for a patch we carry for SDP, and more for being able to upstream it at some point.)

hbase/stackable/patches/2.6.1/0005-Allow-overriding-ipc-bind-port-and-use-alternative-p.patch
```diff
@@ -0,0 +1,195 @@
...
+         clusterConnection.getConnectionMetrics(), Collections.emptyMap());
+       span.setStatus(StatusCode.OK);
+     } catch (Throwable t) {
+@@ -1533,6 +1544,8 @@ public class HRegionServer extends Thread
+     } else {
+       serverLoad.setInfoServerPort(-1);
+     }
++    // serverLoad.setInfoServerPort(Integer.parseInt(System.getenv("HBASE_INFO_PORT")));
```
nightkr, line 129: Dead code

hbase/stackable/patches/2.6.1/0005-Allow-overriding-ipc-bind-port-and-use-alternative-p.patch
adwk67: Have added a flag that defaults to the new behaviour: bd0b75e

hbase/stackable/patches/2.6.1/0005-Allow-overriding-ipc-bind-port-and-use-alternative-p.patch
adwk67: `hbase.info.port `is now `hbase.master.bound.info.port` and `hbase.regionserver.bound.info.port`.
We are not touching `MASTER_INFO_PORT` (= "hbase.master.info.port"): is that you meant with "leave the existing key as the advertised port, introduce a new key for overriding the bound port"?

hbase/stackable/patches/2.6.1/0005-Allow-overriding-ipc-bind-port-and-use-alternative-p.patch
adwk67: de0c6c7

hbase/stackable/patches/2.6.1/0005-Allow-overriding-ipc-bind-port-and-use-alternative-p.patch
```diff
@@ -75,16 +147,16 @@ index 27bcef2f06..80949070fd 100644
...
     * @deprecated since 2.4.0 and will be removed in 4.0.0. Use
     *             {@link HRegionServer#UNSAFE_RS_HOSTNAME_DISABLE_MASTER_REVERSEDNS_KEY} instead.
-@@ -669,6 +674,8 @@ public class HRegionServer extends Thread
+@@ -669,6 +676,8 @@ public class HRegionServer extends Thread
        this.namedQueueRecorder = NamedQueueRecorder.getInstance(this.conf);
        rpcServices = createRpcServices();
        useThisHostnameInstead = getUseThisHostnameInstead(conf);
 +      useThisPortInstead = getUseThisPortInstead(conf);
-+      useThisInfoPortInstead = conf.getInt("hbase.info.port" , this.infoServer != null ? this.infoServer.getPort() : -1);
++      useThisInfoPortInstead = getUseThisInfoPortInstead(conf); //conf.getInt("hbase.info.port" , this.infoServer != null ? this.infoServer.getPort() : -1);
```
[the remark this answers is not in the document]
adwk67: 496444d

hbase/stackable/patches/2.6.1/0005-Allow-overriding-ipc-bind-port-and-use-alternative-p.patch
```diff
@@ -1,49 +1,119 @@
...
++
+   /** Default port region server listens on. */
+   public static final int DEFAULT_REGIONSERVER_PORT = 16020;
+ 
+@@ -325,6 +340,9 @@ public final class HConstants {
+   /** A configuration key for regionserver info port */
+   public static final String REGIONSERVER_INFO_PORT = "hbase.regionserver.info.port";
+ 
++  /** A configuration key for bound regionserver hbase info port */
++  public static final String REGIONSERVER_BOUND_INFO_PORT = "hbase.regionserver.bound.info.port";
```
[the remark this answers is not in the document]
adwk67: 496444d

hbase/stackable/patches/2.6.1/0005-Allow-overriding-ipc-bind-port-and-use-alternative-p.patch
```diff
@@ -1,49 +1,119 @@
...
++  public static final String MASTER_IPC_PORT = "hbase.master.ipc.port";
++
+   /** default port that the master listens on */
+   public static final int DEFAULT_MASTER_PORT = 16000;
+ 
+@@ -206,6 +212,9 @@ public final class HConstants {
+   /** Configuration key for master web API port */
+   public static final String MASTER_INFO_PORT = "hbase.master.info.port";
+ 
++  /** Configuration key for bound master web API port */
```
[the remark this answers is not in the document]
adwk67: 496444d

hbase/stackable/patches/2.6.1/0005-Allow-overriding-ipc-bind-port-and-use-alternative-p.patch
```diff
@@ -1,49 +1,119 @@
...
+--- a/hbase-common/src/main/java/org/apache/hadoop/hbase/HConstants.java
++++ b/hbase-common/src/main/java/org/apache/hadoop/hbase/HConstants.java
+@@ -197,6 +197,12 @@ public final class HConstants {
+   /** Parameter name for port master listens on. */
+   public static final String MASTER_PORT = "hbase.master.port";
+ 
++  /** Parameter name for master IPC address */
++  public static final String MASTER_IPC_ADDRESS = "hbase.master.ipc.address";
++
++  /** Parameter name for master IPC port */
```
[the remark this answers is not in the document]
adwk67: 496444d

hbase/stackable/patches/2.6.1/0005-Allow-overriding-ipc-bind-port-and-use-alternative-p.patch
```diff
@@ -1,49 +1,119 @@
...
++  public static final String MASTER_IPC_PORT = "hbase.master.ipc.port";
++
+   /** default port that the master listens on */
+   public static final int DEFAULT_MASTER_PORT = 16000;
+ 
+@@ -206,6 +212,9 @@ public final class HConstants {
+   /** Configuration key for master web API port */
```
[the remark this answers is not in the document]
adwk67: 496444d

hbase/stackable/patches/2.6.1/0005-Allow-overriding-ipc-bind-port-and-use-alternative-p.patch
adwk67: So
`public static final String MASTER_BOUND_INFO_PORT = "hbase.master.bound.info.port";`
should be renamed to something like:
`public static final String MASTER_ADVERTISED_INFO_PORT = "hbase.master.advertised.info.port";`
as `MASTER_INFO_PORT` is already the bound one?

nightkr: That would be one way, yeah. But as I said above, I think we should try to be consistent about whether the "old" property (`MASTER_INFO_PORT`) is for the bound or advertised port.

adwk67: I've endeavoured to make that change in a separate patch: ee7ea2e

hbase/hbase/stackable/patches/2.6.1/0006-Update-property-usage-for-bound-ports.patch
```diff
@@ -0,0 +1,174 @@
...
++    int port = conf.getInt(REGIONSERVER_INFO_PORT, 0);
+     return port != 0 ? port : this.infoServer != null ? this.infoServer.getPort() : -1;
+   }
+ 
+@@ -2463,11 +2464,11 @@ public class HRegionServer extends Thread
+    */
+   private void putUpWebUI() throws IOException {
+     int port =
+-      this.conf.getInt(HConstants.REGIONSERVER_INFO_PORT, HConstants.DEFAULT_REGIONSERVER_INFOPORT);
++      this.conf.getInt(REGIONSERVER_BOUND_INFO_PORT, HConstants.DEFAULT_REGIONSERVER_INFOPORT);
```
nightkr, line 124: This has to default to the advertised port, so along the lines of `getInt(REGIONSERVER_BOUND_INFO_PORT, getInt(REGIONSERVER_INFO_PORT, DEFAULT_REGIONSERVER_INFOPORT))`.

hbase/hbase/stackable/patches/2.6.1/0006-Update-property-usage-for-bound-ports.patch
```diff
@@ -0,0 +1,174 @@
...
+     String addr = this.conf.get("hbase.regionserver.info.bindAddress", "0.0.0.0");
+ 
+     if (this instanceof HMaster) {
+-      port = conf.getInt(HConstants.MASTER_INFO_PORT, HConstants.DEFAULT_MASTER_INFOPORT);
++      port = conf.getInt(HConstants.MASTER_BOUND_INFO_PORT, HConstants.DEFAULT_MASTER_INFOPORT);
```
nightkr, line 129: (And here)

hbase/hbase/stackable/patches/2.6.1/0006-Update-property-usage-for-bound-ports.patch
```diff
@@ -0,0 +1,174 @@
...
+       reportWebUIPort("master", masterInfoPort);
+-      confMap.put(HConstants.MASTER_INFO_PORT, masterInfoPort);
++      confMap.put(HConstants.MASTER_BOUND_INFO_PORT, masterInfoPort);
+     } else if (serverType == ServerType.RS) {
+       confMap.put(HConstants.REGIONSERVER_PORT, rpcPort);
+ 
+       int rsInfoPort = HBaseTestingUtility.randomFreePort();
+       reportWebUIPort("region server", rsInfoPort);
+-      confMap.put(HConstants.REGIONSERVER_INFO_PORT, rsInfoPort);
++      confMap.put(HConstants.REGIONSERVER_BOUND_INFO_PORT, rsInfoPort);
```
nightkr, line 171: Assuming that we do the fallback, existing tests should be ok as is, I think?

hbase/hbase/stackable/patches/2.6.1/0006-Update-property-usage-for-bound-ports.patch
```diff
@@ -0,0 +1,174 @@
...
+       if (
+-        conf.getInt(HConstants.MASTER_INFO_PORT, 0) != -1
+-          && conf.getInt(HConstants.MASTER_INFO_PORT, HConstants.DEFAULT_MASTER_INFOPORT)
++        conf.getInt(HConstants.MASTER_BOUND_INFO_PORT, 0) != -1
++          && conf.getInt(HConstants.MASTER_BOUND_INFO_PORT, HConstants.DEFAULT_MASTER_INFOPORT)
+               == HConstants.DEFAULT_MASTER_INFOPORT
+       ) {
+         LOG.debug("Setting Master InfoServer Port to random.");
+-        conf.set(HConstants.MASTER_INFO_PORT, "0");
++        conf.set(HConstants.MASTER_BOUND_INFO_PORT, "0");
```
nightkr, line 63: (And here)

hbase/hbase/stackable/patches/2.6.1/0006-Update-property-usage-for-bound-ports.patch
```diff
@@ -0,0 +1,174 @@
...
+         LOG.info("Retry starting http info server with port: " + port);
+       }
+     }
+-    port = useThisInfoPortInstead;
+-    conf.setInt(HConstants.REGIONSERVER_INFO_PORT, port);
++
++    // update bound ports
++    port = this.infoServer.getPort();
++    conf.setInt(REGIONSERVER_BOUND_INFO_PORT, port);
++    conf.setInt(HConstants.MASTER_BOUND_INFO_PORT, port);
```
nightkr, line 143: Is anything actually reading these new values?

hbase/hbase/stackable/patches/2.6.1/0006-Update-property-usage-for-bound-ports.patch
```diff
@@ -0,0 +1,174 @@
...
++
++    // set advertised ports
++    conf.setInt(REGIONSERVER_INFO_PORT, useThisInfoPortInstead);
```
nightkr, line 146: Or this?

hbase/hbase/stackable/bin/hbase-entrypoint.sh
```diff
@@ -72,6 +75,16 @@ if [ -f /stackable/kerberos/krb5.conf ]; then
...
 
+# Service endpoints
+HBASE_SERVICE_PORT=$(cat /stackable/listener/default-address/ports/"${HBASE_PORT_NAME}")
+HBASE_INFO_PORT=$(cat /stackable/listener/default-address/ports/"${HBASE_UI_PORT_NAME}")
+HBASE_LISTENER_ENDPOINT="$HBASE_SERVICE_HOST:$HBASE_INFO_PORT"
+
+sed -i -e s/\$\{HBASE_SERVICE_HOST\}/"${HBASE_SERVICE_HOST}"/g /stackable/conf/hbase-site.xml
+sed -i -e s/\$\{HBASE_SERVICE_PORT\}/"${HBASE_SERVICE_PORT}"/g /stackable/conf/hbase-site.xml
+sed -i -e s/\$\{HBASE_LISTENER_ENDPOINT\}/"${HBASE_LISTENER_ENDPOINT}"/g /stackable/conf/hbase-site.xml
+sed -i -e s/\$\{HBASE_INFO_PORT\}/"${HBASE_INFO_PORT}"/g /stackable/conf/hbase-site.xml
```
nightkr, line 86: Any reason for this to not be using https://github.com/stackabletech/config-utils instead of sed? AFAIU we already pull it in via stackable-base.

nightkr (commented):
A few minor notes, but we're definitely closing in.

If you want to rush it in, we could probably merge 0005 as is, and split 0006 into a separate PR (though that _will_ definitely require coordinating with the operator PR too..).

hbase/hbase/stackable/patches/2.6.1/0006-Update-property-usage-for-bound-ports.patch
adwk67: I'm not 100% sure, but the original code wrote back the port (which may have been incremented in the preceding loop for auto-bind cases) to the conf, so I've retained that. And the original code didn't make a distinction between bound and advertised, so I've written both back.

hbase/hbase/stackable/patches/2.6.1/0006-Update-property-usage-for-bound-ports.patch
adwk67: See comment above.

hbase/hbase/stackable/bin/hbase-entrypoint.sh
adwk67: I've not used config-utils before: not sure how to template the file without first declaring the env-vars. I'll have a look.

hbase/hbase/stackable/patches/2.6.1/0006-Update-property-usage-for-bound-ports.patch
adwk67: That was changed inadvertently - have reverted it: 65aaae2

hbase/hbase/stackable/patches/2.6.1/0006-Update-property-usage-for-bound-ports.patch
adwk67: That was changed inadvertently - have reverted it: 65aaae2

hbase/hbase/stackable/patches/2.6.1/0006-Update-property-usage-for-bound-ports.patch
adwk67: 65aaae2

hbase/hbase/stackable/bin/hbase-entrypoint.sh
adwk67: Will defer this to https://github.com/stackabletech/docker-images/issues/1191

labrenbe (approved):

PR #1164, opened 2025-06-12
chore: Change default user id and gid

lfrancke:
# Description
_Part of https://github.com/stackabletech/issues/issues/645_
_Part of https://github.com/stackabletech/issues/issues/651_

> [!NOTE]  
> This was already [applied once before](https://github.com/stackabletech/docker-images/pull/916) and later [reverted](https://github.com/stackabletech/docker-images/pull/1115).
> It was merged prematurely back then but now the immediate [parent issue](https://github.com/stackabletech/issues/issues/651) has been done and it should be good to go again

Change default user id and gid to the same ones we use in the operators themselves.
These are only the defaults that are used when a Pod does not specify their own `securityContext.runAsUser` or `runAsGroup`.

All of our operators have been changed to not set the user and group id anymore which means they'll fall back to the values from the image.

<details>

```shell
```

</details>


lfrancke:
I have tested this by installing the waterlevel demo on 25.3, updated everything to current 0.0.0-dev and then updated to use custom built images with this PR included.

All pods and tools came up without a problem. The FSGroup setting makes it possible that all mounts are still readable/writable.

sbernauer (approved):
Thanks for the tests!

Issue #1068, opened 2025-04-29
Enable custom versions for patched components


dervoeti:
Apply a standardized SDP version suffix (3.4.1-stackable25.3.0) to all SDP artifacts that include custom modifications. This applies to JARs, Go binaries, Python packages, container images, etc.

Ensure vulnerability scanners continue detecting the original upstream version (e.g., 3.4.1 for Hadoop). This may require adding metadata, dummy components, or another approach.

dervoeti:
Status update: It's working, but custom versions in tools can cause unexpected issues with clients, it's a breaking change. Another idea is that we have custom versions for the libraries but the binaries will have the original version, even though it's the same code. But that means we have to compile some parts of the product twice and it complicates the build process.
Another idea would be to patch the code so that it identifies to clients as the original version even though the JARs have the custom version.
We'll talk about this in a few days and decide on a way forward.

dervoeti:
Implementation complete, we have custom versions for all products that we build from source (everything except Airflow and Superset). We decided to accept potential issues clients might have with the new versioning scheme, since these cases are expected to be rare.

### Release Note
All products that are built from source in SDP, which is at the time of this release everything except Airflow and Superset, now have a version suffix to indicate they include custom modifications made by Stackable. The suffix the same as for SDP container images, so Zookeeper 3.9.3 in SDP 25.7 will for example report `3.9.3-stackable25.7.0` as its version.

PR #1166, opened 2025-06-13
feat: custom product versions

dervoeti:
# Description

Implemented solution for https://github.com/stackabletech/decisions/issues/44
Fixes https://github.com/stackabletech/docker-images/issues/1068

This PR enables custom versions (with a suffix like `-stackable0.0.0-dev`) for all components we apply custom patches to. It also makes the following products use our patched Hadoop libraries:

-   HBase
-   Phoenix
-   hbase-operator-tools
-   Hive
-   Druid
-   Spark

And these components use our patched HBase libraries now:

-   Phoenix
-   hbase-operator-tools
-   Spark

In the SBOMs, the custom version is replaced by the original one (using `sed`), so vulnerabilities filed directly against e.g. Hadoop 3.3.6 are still found when scanning the SBOMs (which we do in our vulnerability management pipeline). Otherwise vulnerabilities against Hadoop 3.3.6 might be missed, because the version would be something like 3.3.6-stackable25.7.0 and vulnerability databases don't contain entries for that particular version.The patched libraries are used by overriding for instance the Hadoop version when running Maven (like `-Dhadoop.compile.version=3.3.6-stackable0.0.0-dev` ). Since they are not found in our Nexus Maven repo, we copy them to the local Maven repo from the Hadoop builder:
```
cp -r /stackable/patched-libs/maven/* /stackable/.m2/repository
```

We can't `COPY` directly into `/stackable/.m2/repository` since that directory is usually cached and the copied libraries would be overridden by the cache.

Further changes:
-   The patch that fixes CVE-2023-34455 in Druid was removed, since Druid now uses the patched Hadoop version which does not contain the vulnerability
-   [HADOOP-18837](https://issues.apache.org/jira/browse/HADOOP-18837) and [HADOOP-18496](https://issues.apache.org/jira/browse/HADOOP-18496) where backported to Hadoop 3.3.6 to check if the vulnerabilities fixed by these patches are also fixed in products that use our patched Hadoop version
-   Building HBase now happens in a separate Dockerfile (`hbase/hbase/Dockerfile`) to differentiate the final HBase image (`hbase/Dockerfile`) from HBase the application. This was needed, because for example Phoenix depends on our patched HBase version now, but the HBase image depends on Phoenix, which would be a circular dependency if the HBase container image and application were actually the same image. Now all components of the HBase image (except `hadoop-s3-builder` ) are built in separate Dockerfiles instead of inline.
-   The same thing for Trino: `trino/trino/Dockerfile` is the application now, since `storage-connector` uses our patched Trino version now, while the Trino image includes the `storage-connector`
-   hbase-operator-tools and Phoenix now have the HBase version they were built with as suffix in their Docker target names (but not in their versions). The reason is, that while the Phoenix version itself is the same (`5.2.1-stackable25.7.0`), in our case Phoenix can be built against HBase 2.6.1 or HBase 2.6.2, which also both introduce a different version of Hadoop. We don't want e.g. the HBase 2.6.2 image include a static Phoenix 5.2.1 version that was built with HBase 2.6.1 and Hadoop 3.3.6. So when we include Phoenix in the HBase image, we need to specify which variant of 5.2.1 we want to include. Same for hbase-operator-tools.

I successfully built these products and tested them with the smoke tests:

-   Kafka 3.9.0
-   Nifi 1.28.1
-   Nifi 2.4.0
-   Hive 3.1.3
-   Hive 4.0.1
-   OPA 1.4.2
-   Zookeeper 3.9.3
-   Druid 30.0.1
-   HBase 2.6.1
-   HBase 2.6.2
-   Hadoop 3.3.6
-   Hadoop 3.4.1
-   Spark-k8s 3.5.5
-   Trino 470
-   Trino 476

I also built omid 1.1.3 successfully, we don't have a smoke test for it as far as I know.

> [!NOTE]

<details>

```shell
```

</details>


dervoeti:
This PR is too large. I will try to split it up.

PR #1167, opened 2025-06-13
chore(krb5): Update readme

NickLarsenNZ:
The image should not be part of the SDP platform, so I recommend we move it to the correct Harbour project, and rename the image so indicate its usage.

> [!CAUTION]
> Update image regerences in ad-init and hdfs-topology-provider. Be sure to use Harbor and the correct project (stackable instead of sdp).

> [!NOTE]

<details>

```shell
```

</details>


NickLarsenNZ:
The image builds:

```
 => => exporting layers                                                                                                                                                                                                                                            1.2s
 => => writing image sha256:663584d2fea34c2aa6cce1052521b1eab2bab59118700d63ef521bd59f9c6519                                                                                                                                                                       0.0s 
 => => naming to oci.stackable.tech/sdp/krb5-testing-tools:1.21.1-stackable0.0.0-dev
```

nightkr:
`krb5-testing-server` maybe? Traditionally we've reserved `-tools` for "toolbox images" rather than daemons.

NickLarsenNZ:
> `krb5-testing-server` maybe? Traditionally we've reserved `-tools` for "toolbox images" rather than daemons.

I was going to call it `-server` until I saw it being used as a client in a container.

I don't mind though.

nightkr:
Ah, right. For some reason I thought we set `CMD`. We don't, so `-tools` should be fine after all ("server is one role it can take, but not the default").

krb5-testing-tools/README.md
```diff
@@ -0,0 +1,13 @@
+# krb5-testing-tools
+
+> [!CAUTION]
+> This image is only intended for Stackable-internal tests.
```
nightkr, line 4: We use it in demos too, right?

nightkr (approved):

krb5-testing-tools/README.md
NickLarsenNZ: Ah yeah... so it's name is kind of bad.
We'll think of an alternative name - otherwise I'll drop the PR.

NickLarsenNZ:
@sbernauer and I chatted about potential naming this morning, as it is used in:
- tests
- demos

For:
- non-prod KDC's
- krb5 client tools

Considering that, we feel `krb5-testing-tools` isn't exactly correct. I did consider dropping the PR, but I still stand by my original reasoning that `krb5` on its own misses some signals about what it is and were it could be used (or shouldn't be used - which, sure, is explained in the docs I updated here)

PR #1171, opened 2025-06-13
feat: custom version for Kafka

dervoeti:
# Description

Part of https://github.com/stackabletech/docker-images/issues/1068

This PR enables custom versions (with a suffix like `-stackable0.0.0-dev`) for Kafka.

In the SBOM, the custom version is replaced by the original one (using `sed`), so vulnerabilities filed directly against Kafka are still detected by scanners when scanning the SBOM.

I built Kafka 3.9.0, tested it with the smoke test and manually validated that it reported the correct version.

> [!NOTE]

<details>

```shell
```

</details>

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -129,6 +129,7 @@ All notable changes to this project will be documented in this file.
 - kafka: Remove `3.7.1` and `3.8.0` ([#1117]).
 - spark-connect-client: Remove `3.5.5` ([#1142]).
 - kafka: Enable custom versions ([#1171]).
+- omid: Enable custom versions ([#1174]).
 
 [nifi-iceberg-bundle]: https://github.com/stackabletech/nifi-iceberg-bundle
 [#1025]: https://github.com/stackabletech/docker-images/pull/1025
@@ -196,6 +197,7 @@ All notable changes to this project will be documented in this file.
 [#1170]: https://github.com/stackabletech/docker-images/pull/1170
 [#1171]: https://github.com/stackabletech/docker-images/pull/1171
 [#1173]: https://github.com/stackabletech/docker-images/pull/1173
+[#1174]: https://github.com/stackabletech/docker-images/pull/1174
 [#1179]: https://github.com/stackabletech/docker-images/pull/1179
 [#1180]: https://github.com/stackabletech/docker-images/pull/1180
 
--- a/omid/Dockerfile
+++ b/omid/Dockerfile
@@ -4,6 +4,7 @@
 FROM stackable/image/java-devel AS builder
 
 ARG PRODUCT
+ARG RELEASE
 ARG DELETE_CACHES="true"
 ARG STACKABLE_USER_UID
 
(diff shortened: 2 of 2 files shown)


kafka/Dockerfile
```diff
@@ -21,22 +22,34 @@ COPY --chown=${STACKABLE_USER_UID}:0 kafka/stackable/patches/${PRODUCT} /stackab
 RUN <<EOF
 cd "$(/stackable/patchable --images-repo-root=src checkout kafka ${PRODUCT})"
 
+ORIGINAL_VERSION="${PRODUCT}"
+NEW_VERSION="${PRODUCT}-stackable${RELEASE}"
+
+# The comment in gradle.properties mentions to update the version in all of these files
```
razvan, line 28: The sed commands don't fail if there is nothing to do.

```bash
$ sed -i 's/razvan/mihai/g' gradle.properties
$ echo $?
0
```
Instead of relying on this comment I did a full search and replace for 3.9.1
```bash
$ fd --type file | xargs perl -pi -e 's/\b3\.9\.1\b/3.9.1-stackable/g'
$ git status
HEAD detached at 3.9.1
Changes not staged for commit:
  (use "git add <file>..." to update what will be committed)
  (use "git restore <file>..." to discard changes in working directory)
        modified:   docs/configuration.html
        modified:   docs/js/templateData.js
        modified:   docs/upgrade.html
        modified:   gradle.properties
        modified:   streams/quickstart/java/pom.xml
        modified:   streams/quickstart/java/src/main/resources/archetype-resources/pom.xml
        modified:   streams/quickstart/pom.xml
        modified:   tests/kafkatest/__init__.py
        modified:   tests/kafkatest/version.py
```
It differs from the `sed` commands in that:
1. The DEV_VERSION in `tests/kafkatest/version.py` retains the SNAPSHOT qualifier.
2. There is no change in `kafka-merge-pr.py`. It turns out that file contains the _wrong_ version already (`3.9.0.`)

razvan (commented):
Have you thought of any possibility to integrate the version patching in patchable?

This would ensure reproducibility outside the docker build and allow for a dev to check that the version update is actually correct.

dervoeti:
> Have you thought of any possibility to integrate the version patching in patchable?
> 
> This would ensure reproducibility outside the docker build and allow for a dev to check that the version update is actually correct.

I thought about that as well, but didn't find a good solution yet. But now that I think about it again... I could create a patch file that appends `-stackable0.0.0-dev` to all the versions and then replace that placeholder with `-stackable${RELEASE}` before building Kafka. Maybe that's better, it would still require replacing before running Gradle, but could just replace `-stackable0.0.0-dev` across all files. So at least we'd have separate patch files for each Kafka version which adapt the original version in different source files (which files exactly can depend on the Kafka version), that would be an improvement. Currently the files that get changed are static for all Kafka versions. And you get the final source code for the dev build when running `patchable checkout`.
I could implement that, I think it would improve things.

dervoeti:
> > I could implement that, I think it would improve things.
> 
> So do you still want this PR merged as is first?

No, I'll adapt it later today

dervoeti:
@razvan I updated the process now, seems to work

razvan (approved):
lgtm assuming the hive patch in this pr was intentional :)

PR #1173, opened 2025-06-13
feat: custom product versions for Hadoop, HBase, Phoenix, hbase-operator-tools, Druid, Hive and Spark

dervoeti:
# Description

Part of https://github.com/stackabletech/docker-images/issues/1068

This PR enables custom versions (with a suffix like `-stackable0.0.0-dev`) for components we apply custom patches to. It also makes the following products use our patched Hadoop libraries:

-   HBase
-   Phoenix
-   hbase-operator-tools
-   Hive
-   Druid
-   Spark

And these components use our patched HBase libraries now:

-   Phoenix
-   hbase-operator-tools
-   Spark

In the SBOMs, the custom version is replaced by the original one (using `sed`), so vulnerabilities filed directly against e.g. Hadoop 3.3.6 are still found when scanning the SBOMs (which we do in our vulnerability management pipeline). Otherwise vulnerabilities against Hadoop 3.3.6 might be missed, because the version would be something like 3.3.6-stackable25.7.0 and vulnerability databases don't contain entries for that particular version.The patched libraries are used by overriding for instance the Hadoop version when running Maven (like `-Dhadoop.compile.version=3.3.6-stackable0.0.0-dev` ). Since they are not found in our Nexus Maven repo, we copy them to the local Maven repo from the Hadoop builder:
```
cp -r /stackable/patched-libs/maven/* /stackable/.m2/repository
```

We can't `COPY` directly into `/stackable/.m2/repository` since that directory is usually cached and the copied libraries would be overridden by the cache.

Further changes:
-   The patch that fixes CVE-2023-34455 in Druid was removed, since Druid now uses the patched Hadoop version which does not contain the vulnerability
-   Building HBase now happens in a separate Dockerfile (`hbase/hbase/Dockerfile`) to differentiate the final HBase image (`hbase/Dockerfile`) from HBase the application. This was needed, because for example Phoenix depends on our patched HBase version now, but the HBase image depends on Phoenix, which would be a circular dependency if the HBase container image and application were actually the same image. Now all components of the HBase image (except `hadoop-s3-builder` ) are built in separate Dockerfiles instead of inline.
-   hbase-operator-tools and Phoenix now have the HBase version they were built with as suffix in their Docker target names (but not in their versions). The reason is, that while the Phoenix version itself is the same (`5.2.1-stackable25.7.0`), in our case Phoenix can be built against HBase 2.6.1 or HBase 2.6.2, which also both introduce a different version of Hadoop. We don't want e.g. the HBase 2.6.2 image include a static Phoenix 5.2.1 version that was built with HBase 2.6.1 and Hadoop 3.3.6. So when we include Phoenix in the HBase image, we need to specify which variant of 5.2.1 we want to include. Same for hbase-operator-tools.

I successfully built these products and tested them with the smoke tests:

-   Hive 3.1.3
-   Hive 4.0.1
-   Druid 30.0.1
-   HBase 2.6.1
-   HBase 2.6.2
-   Hadoop 3.3.6
-   Hadoop 3.4.1
-   Spark-k8s 3.5.5

> [!NOTE]

<details>

```shell
```

</details>


hadoop/Dockerfile
```diff
@@ -74,27 +76,42 @@ COPY --chown=${STACKABLE_USER_UID}:0 hadoop/stackable/jmx /stackable/jmx
...
-tar -czf /stackable/hadoop-${PRODUCT}-src.tar.gz .
+tar -czf /stackable/hadoop-${NEW_VERSION}-src.tar.gz .
 
 mvn \
     --batch-mode \
     --no-transfer-progress \
-    clean package \
+    clean package install \
     -Pdist,native \
-    -pl '!hadoop-tools/hadoop-pipes,!hadoop-yarn-project,!hadoop-mapreduce-project,!hadoop-minicluster' \
```
adwk67, line 85: We delete these jars later on anyway, but I think these were omitted here to speed up the build. Do we need to re-include them?

druid/stackable/patches/30.0.1/0010-Fix-CVE-2023-34455.patch
```diff
@@ -1,43 +0,0 @@
...
-+                            <execution>
-+                                <id>fix-cve-2023-34455-remove-snappy</id>
-+                                <phase>package</phase>
-+                                <goals>
-+                                    <goal>exec</goal>
-+                                </goals>
-+                                <configuration>
-+                                  <executable>/usr/bin/rm</executable>
-+                                    <arguments>
-+                                      <argument>${project.build.directory}/hadoop-dependencies/hadoop-client-api/3.3.6/snappy-java-1.1.8.2.jar</argument>
```
adwk67, line 36: I assumed this was removed here as it was dealt with in the hadoop build, but I don't seen a patch for it there?

hive/Dockerfile
```diff
@@ -39,26 +42,33 @@ RUN --mount=type=cache,id=maven-hive-${PRODUCT},uid=${STACKABLE_USER_UID},target
 BUILD_SRC_DIR="$(/stackable/patchable --images-repo-root=src checkout hive ${PRODUCT})"
 cd "$BUILD_SRC_DIR"
 
+# Make Maven aware of custom Stackable libraries
+cp -r /stackable/patched-libs/maven/* /stackable/.m2/repository
+
+ORIGINAL_VERSION=$(mvn help:evaluate -Dexpression=project.version -q -DforceStdout)
```
adwk67, line 48: is this used at all?

spark-k8s/Dockerfile
```diff
@@ -55,8 +61,13 @@ COPY --chown=${STACKABLE_USER_UID}:0 spark-k8s/hbase-connectors/stackable/patche
 RUN <<EOF
 cd "$(/stackable/patchable --images-repo-root=src checkout spark-k8s/hbase-connectors ${HBASE_CONNECTOR})/spark"
 
+ORIGINAL_VERSION=$(mvn help:evaluate -Dexpression=project.version -q -DforceStdout)
```
adwk67, line 64: is this value used? (it is re-defined further down)

druid/stackable/patches/30.0.1/0010-Fix-CVE-2023-34455.patch
dervoeti: Yes it's patched in Hadoop 3.3.6, this is the patch file: https://github.com/stackabletech/docker-images/blob/main/hadoop/stackable/patches/3.3.6/0007-Bump-Snappy-version-to-fix-CVEs.patch

spark-k8s/Dockerfile
dervoeti: Wasn't used, removed it

adwk67 (approved):

PR #1202, opened 2025-07-11
fix: Change GID to 1000

Techassi:
#1193 changed the UID to 1000, but the GID was set to 0. This caused issues, because the root group with ID 0 is already present. As such, the `groupadd` command fails which in turn breaks _all_ image builds.

**This PR set the GID to 1000 as well.**

Local test builds succeeded:

- stackable-base@1.0.0
- opa@1.4.2


Techassi:
Ha! That's true. I will remove it.

PR #1203, opened 2025-07-11
fix: Use STACKABLE_USER_GID arg instead of the hard-coded GID of 0

NickLarsenNZ:
Images were failing to build, and we think it is due to the GID being `1000`, but permissions being set to GID `0`.

Error:

```
failed to solve: failed to compute cache key: failed to calculate checksum of ref sx27n0fucs3yx7dk9rffa305j::0l8bk4seqdjwhzu4loqedvxlm: "/stackable/patched-libs": not found
```

Maybe because of:

```sh
```

[Slack thread](https://stackable-workspace.slack.com/archives/C07UG6JH44F/p1752243729159399)

### DIFF
--- a/airflow/Dockerfile
+++ b/airflow/Dockerfile
@@ -44,6 +44,7 @@ ARG SHARED_STATSD_EXPORTER
 ARG PYTHON
 ARG TARGETARCH
 ARG STACKABLE_USER_UID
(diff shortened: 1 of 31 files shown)


NickLarsenNZ:
Hmm, well I'm stumped.

PR #1219, opened 2025-07-24
Use our build-repo to cache NPM dependencies

lfrancke:
# Description

We're already using our mirror to cache Maven dependencies.
This uses the same mirror for NPM dependencies.

I tried building all of these and they worked:

- [x] Airflow
- [x] Druid
- [x] HBase
- [x] Hadoop HDFS
- [x] Hive
- [x] Kafka
- [x] NiFi
- [x] Spark
- [x] Superset
- [x] Trino
- [x] ZooKeeper
- [x] OPA

## How I tested this
I did apply the patch and built all images in all versions. They all succeeded so I'm sure that things have not gotten worse. At the same time I tailed the logs for our build-repo server and saw accesses to the NPM mirror. As we're not using that mirror anywhere else I'm fairly certain that these requests must be coming from me.

As a side-effect I also built a spike which introduces a proxy in our build process which can intercept these calls. The branch is available here https://github.com/stackabletech/docker-images/tree/feat/proxy but I did not test it with this yet. That's for the future. This is purely informational and this branch is NOT READY for use.

After these tests I'm convinced that the mirroring will work for at least some projects and it does not break any of them.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,12 @@ All notable changes to this project will be documented in this file.
 
 ## [Unreleased]
 
+### Changed
+
+- all: Use our build-repo to cache NPM dependencies ([#1219])
+
+[#1219]: https://github.com/stackabletech/docker-images/pull/1219
+
 ## [25.7.0] - 2025-07-23
 
 ## [25.7.0-rc1] - 2025-07-18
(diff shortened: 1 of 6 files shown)


maltesander (approved):
LGTM!

Issue #1181, opened 2025-06-18
images: replace UBI maven package with newer version


razvan:
# Description

The UBI Maven package installs a Maven version that is too old for some projects to build with.
Because of this, different products (Nifi) install different versions of Maven and sometimes (Spark) use different versions in the same image.

This makes products build inconsistent and even increases built times.

# Proposed Solution

Replace the UBI Maven package in the `java-devel` with the repo.stackable.tech version and ensure all products use it exclusively.

# Size

S

PR #1220, opened 2025-07-26
Use custom Maven version for all our products

lfrancke:
# Description

/fixes https://github.com/stackabletech/docker-images/issues/1181

Maven as of 3.9+ supports an environment variable `MAVEN_ARGS` to standardise certain arguments to be used for every Maven invocation. This allows us to globally silence the transfer progress and enable batch mode and it allows us to remove the workarounds we had in place for Spark, Trino and NiFi.

## How this was tested

I manually built all images and they succeeded. I did not run any integration tests or anything but I believe that once they built successfully it should be good.


lfrancke:
Thank you! I indeed forgot about that.

PR #1228, opened 2025-08-29
Remove the demo configuration of the OpenSearch Security plugin

siegfriedweber:
# Description

Remove the demo configuration of the OpenSearch Security plugin and replace it with a minimal configuration

The demo configuration contains – among other things – pre-defined users with well-known credentials. If the security index is initialized with `plugins.security.allow_default_init_securityindex: "true"` and not all files of the configuration are correctly overriden by `podOverrides`, then the cluster is vulnerable. Therefore, this configuration is replaced with a minimal one. OpenSearch starts successfully with this minimal configuration but users and authentication have to be defined explicitly to use the cluster.

Part of stackabletech/opensearch-operator#2

> [!NOTE]

<details>

```shell
```

</details>


labrenbe (approved):
lgtm

PR #1288, opened 2025-09-29
fix(boil): Revert accidental default registry change from #1281

Techassi:
This was accidentally changed from `oci.stackable.tech` to `localhost`.


adwk67 (approved):
thanks for fixing this! This confused me earlier today :) (tested by building in this branch and comparing to build from main)

PR #1299, opened 2025-10-08
ci(build): Add runners selection

Techassi:
> [!CAUTION]
> Currently untested. A test run will be triggered to confirm this change works as expected.

With this PR, container image build pipelines can now select one of the following set of runners:

- `mixed`: x86_64 builds run on the GitHub hosted runners and the aarch64 builds run on the Ubicloud runners
- `ubicloud`: Both the x86_64 and the aarch64 builds run on the Ubicloud runners

This PR is raised to fix the OpenSearch Dashboards build on x86_64. It is very likely that this feature will be removed or adjusted in the future when the set of available/used runners changes.

### DIFF
--- a/.github/workflows/reusable_build_image.yaml
+++ b/.github/workflows/reusable_build_image.yaml
@@ -10,6 +10,13 @@ on:
       registry-namespace:
         required: true
         type: string
+      runners:
+        description: |
+          The type of runners used.
+          - `mixed`: x86_64 builds run on the GitHub hosted runners and the aarch64 builds run on the Ubicloud runners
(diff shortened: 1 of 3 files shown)


Techassi:
> LGTM, probably also need to update [`main`/.github/actionlint.yaml](https://github.com/stackabletech/docker-images/blob/main/.github/actionlint.yaml?rgh-link-date=2025-10-08T09%3A28%3A17Z)

Huh, that file would indeed need updating, but we are already using a runner which is not listed in there and it seems like there are no complaints. I assume this is because we use the matrix to set the `runs-on` field and `actionlint` might not be smart enough to pick this up.

I can update the file anyway.

PR #1310, opened 2025-10-16
fix(nifi): NiFi OPA plugin path in final image

labrenbe:
# Description

Copy the plugin into `lib` directory instead of `extensions` and use a standardized name for the nar file. 

> [!NOTE]

<details>

```shell
```

</details>


labrenbe:
> LGTM, I assume it builds locally.
> 
> From memory, this removes a warning on startup. Is there any other effect?

No, the only other change is that the name of the nar-file will follow the naming convention.

PR #1325, opened 2025-11-04
fix(nifi): Do not require SNI checks

NickLarsenNZ:
This was already done for 2.4.0 in https://github.com/stackabletech/docker-images/pull/1265

> [!TIP]
> No, this isn't going against agreed upon values. We did have a discussion about setting the Host header check to false, but "secure by default" was the reason for keeping that enabled and documenting how to disable it.


xeniape (approved):

sbernauer:
Patch me baby one more one more one more time

PR #1331, opened 2025-11-05
ci(trino): Use Ubicloud runners only

Techassi:
Since building Vector from source (see #1323), we seem to run out of disk space on the included GitHub runners. As such, we now only use the larger Ubicloud runners to build Trino.

Test run: https://github.com/stackabletech/docker-images/actions/runs/19102981903

### DIFF
--- a/.github/workflows/build_trino.yaml
+++ b/.github/workflows/build_trino.yaml
@@ -35,5 +35,6 @@ jobs:
       product-name: trino
       sdp-version: ${{ github.ref_type == 'tag' && github.ref_name || '0.0.0-dev' }}
       registry-namespace: sdp
-      # We have a too high disk usage for GitHub runners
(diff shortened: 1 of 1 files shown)


.github/workflows/build_trino.yaml
```diff
@@ -35,3 +35,4 @@ jobs:
       product-name: trino
       sdp-version: ${{ github.ref_type == 'tag' && github.ref_name || '0.0.0-dev' }}
       registry-namespace: sdp
+      runners: ubicloud
```
sbernauer: I think it would be great if you could add a short comment on the motivation of using ubicloud here
```suggestion
      # We have a too high disk usage for GitHub runners
      runners: ubicloud
```

sbernauer (approved):

PR #1350, opened 2025-11-19
chore: Deny unwraps and panics in Rust code

Techassi:
Cloudflare's outage caused by an `unwrap` triggered me to start tightening the use of `unwrap`s in our code bases. This will be the first PR in a series of PR across various of our repositories which use Rust.

I chose this repository as an initial playground, because the Rust usage is fairly limited and any potential fallout (having to deal with existing `unwrap`s) is kept to a minimum.


NickLarsenNZ (approved):
LGTM

PR #1398, opened 2026-01-15
ci(opensearch): Use Ubicloud runners

Techassi:
Needed because OpenSearch doesn't build on the GH hosted runner anymore because of missing disk space (bloat).

Successful test run: https://github.com/stackabletech/docker-images/actions/runs/21028162566


NickLarsenNZ (approved):
LGTM

PR #1399, opened 2026-01-15
ci(hive): Use Ubicloud runners

Techassi:
The image started to fail to build on January 14th 2026.

We are not exactly sure why the builds now fail due to missing disk space, because Hive itself wasn't changed. It might be because of other changes, like the Vector bump.

Successful est run: https://github.com/stackabletech/docker-images/actions/runs/21028434262


NickLarsenNZ (approved):
LGTM

PR #1456, opened 2026-04-07
fix(superset): Patch pinned packages and re-generate lock file

Techassi:
This PR pins the `@types/offscreencanvas` package of Superset (frontend) 4.1.4 to 2019.7.0 in order to fix the following issue encountered since the last few scheduled CI runs. Additionally, it adds a patch with a re-generated package-lock.json file which fully fixes the following error:

```
Invalid: lock file's @types/offscreencanvas@2019.7.0 does not satisfy @types/offscreencanvas@2019.7.3
```

This error is very similar to errors fixed in #1363, #1315, and #1316. This is a known weakness/bug in npm.

Test builds in CI:

- https://github.com/stackabletech/docker-images/actions/runs/24068907048 (failed)
- https://github.com/stackabletech/docker-images/actions/runs/24134252438 (succeeded)


Techassi:
Okay one image build succeeded, one failed because of an unsuccessful SSL/TLS handshake. Re-running.

NickLarsenNZ (approved):
LGTM, but I would recommend using the container image (also keeps the Dockerfile simpler)

Issue #1465, opened 2026-04-16
chore: Faster image builds


NickLarsenNZ:
Many images rebuild the same thing as a stage (for example, hadoop, druid, etc... all build hadoop/hadoop, then copy things out in the final image).

One way we can speed up image builds is move that compile stage out on its own, and publish the image for reuse.

Each version would only need to be built once (per architecture, considering platform dependent code). Any patches applied after that would cause a new build. The builds would be timestamped.

> [!IMPORTANT]
> The SDP version is irrelevant to the source build, so the same artifact could be used for 0.0.0-dev images and the actual SDP release images.
> We decided to abuse the `sdp-version` input of the reusable workflow to store a timestamp, so the source images will be `oci.stackable.tech/precompiled/hadoop:1.2.3-stackable1234567`.
> - The timestamp makes it easier to see which is the newer build.
> - A second PR can be raised to update images to use the new timestamp.
> - If that becomes a burden, we can explore tooling to make this easier (maybe in boil).

PRs will be raised in pairs. For example:

1. PR for `precompiled/hadoop` (or vector, or nifi) so the image is available.
2. PR to replace the hadoop-builder (or vector, or nifi) stage for the dependent product images.

## Groups of changes

We don't _have_ to do all at once, but instead we can do the ones that appear to take significant time.

### Hadoop

> [!CAUTION]
> This is currently blocked by how we handle the visible version string in the Hadoop UI as well as considering it being consistent with what appears in the SBOM.
> See the [comment](https://github.com/stackabletech/docker-images/pull/1472#issuecomment-4287943665) in the hadoop PR for more details

- https://github.com/stackabletech/docker-images/pull/1466
- https://github.com/stackabletech/docker-images/pull/1472
- druid: TODO
- ...

### Vector

> [!NOTE]
> This will be somewhat quicker since #1462

- precompiled/vector: TODO
- whatever uses vector: TODO

### Nifi

- precompiled/nifi: TODO
- nifi: TODO

## Example of the Part 2 type changes

<details>
<summary>Sample of the second part of this change</summary>

### `hadoop/boil.toml`

```diff
diff --git a/hadoop/boil-config.toml b/hadoop/boil-config.toml
index 4c6ef2a..a7989da 100644
--- a/hadoop/boil-config.toml
+++ b/hadoop/boil-config.toml
@@ -3,21 +3,27 @@
 
 # Not part of SDP 25.7.0, but still required for hbase, hive, spark-k8s
[...]

PR #1466, opened 2026-04-16
chore(precompiled/hadoop): Add image for precompiled hadoop for later reuse

NickLarsenNZ:
Part of https://github.com/stackabletech/docker-images/issues/1465

Compile hadoop and publish is as an image that can later be used to replace the `hadoop builder` stage in hadoop/Dockerfile.

In the second part of this change, we will remove hadoop/hadoop, and use precompiled/hadoop.

### To the reviewer

You can see the hadoop/hadoop source was moved verbatim in 83234916c75593c3a7e13e0106777e040f643724, and then references to hadoop/hadoop were updated to precompiled/hadoop in bf2453ca98166055e097bf8604b0994a1f75d7cb.

See the parent issue to understand reasoning for things that might look unusual.


sbernauer (approved):
WCGW?

PR #1468, opened 2026-04-16
ci(patchable): Build on PR

Techassi:
Previous PR #1459 silently broke the patchable build, because we currently do not build patchable on PRs (like we do for boil).


NickLarsenNZ (approved):
LGTM

PR #1469, opened 2026-04-16
feat(boil): Relax vendor version constraint

NickLarsenNZ:
Previously we required the vendor version to be a valid semver (like we use for SDP releases: 26.3.0). This constraint has been relaxed to allow any valid container tag characters.

Also replaced `semver` crate with `regex`.

## Example of misuse

```
❯ cargo boil some-chicken hadoop=3.4.2 --image-version .
error: invalid value '.' for '--image-version <IMAGE_VERSION>': invalid image tag characters for "."
```


rust/boil/src/cli/build.rs
```diff
@@ -129,16 +130,29 @@ pub struct BuildArguments {
...
+    LazyLock::new(|| Regex::new(r"^[a-zA-Z0-9_][a-zA-Z0-9_.-]{0,127}$").expect("regex is valid"));
+
 impl BuildArguments {
-    fn parse_image_version(input: &str) -> Result<Version, ParseImageVersionError> {
-        let version = Version::from_str(input).context(ParseVersionSnafu)?;
-        ensure!(version.build.is_empty(), ContainsBuildMetadataSnafu);
+    /// Ensure that the given version will be valid for use in the image tag
+    fn parse_image_version(version: &str) -> Result<String, ParseImageVersionError> {
+        // TODO (@NickLarsenNZ): Probably want to minus the characters used in front (x.y.z-stackable).
+        // But maybe that validation needs to go up a layer.
```
NickLarsenNZ: Pointing out this todo. Please recommend something.

rust/boil/src/cli/build.rs
```diff
@@ -129,16 +130,29 @@ pub struct BuildArguments {
     pub rest: Vec<String>,
 }
 
+// This is derived from the general rule where the length of the tag can be up to 128 chars
+// But that checking needs to be at a higher layer.
+static VALID_IMAGE_TAG: LazyLock<Regex> =
+    LazyLock::new(|| Regex::new(r"^[a-zA-Z0-9_][a-zA-Z0-9_.-]{0,127}$").expect("regex is valid"));
```
NickLarsenNZ: Pointing out this todo. Please recommend something.

NickLarsenNZ: I could drop the length checking here...


```suggestion
// This is derived from the general rule where the length of the tag can be up to 128 chars
// But that checking needs to be at a higher layer.
static VALID_IMAGE_TAG: LazyLock<Regex> =
    LazyLock::new(|| Regex::new(r"^[a-zA-Z0-9_][a-zA-Z0-9_.-]+").expect("regex is valid"));
```

rust/boil/src/cli/build.rs
NickLarsenNZ: Per below, we can remove length checking here

```suggestion
```

rust/boil/src/cli/build.rs
```diff
@@ -129,16 +130,29 @@ pub struct BuildArguments {
...
-        let version = Version::from_str(input).context(ParseVersionSnafu)?;
-        ensure!(version.build.is_empty(), ContainsBuildMetadataSnafu);
+    /// Ensure that the given version will be valid for use in the image tag
+    fn parse_image_version(version: &str) -> Result<String, ParseImageVersionError> {
+        // TODO (@NickLarsenNZ): Probably want to minus the characters used in front (x.y.z-stackable).
+        // But maybe that validation needs to go up a layer.
+        if version.len() > 128 {
+            return VersionTooLongSnafu.fail();
+        }
+
```
NickLarsenNZ: We can remove length checking from here

```suggestion
```

rust/boil/src/cli/build.rs
```diff
@@ -160,11 +174,11 @@ impl BuildArguments {
 
 #[derive(Debug, Snafu)]
 pub enum ParseImageVersionError {
-    #[snafu(display("failed to parse semantic version"))]
-    ParseVersion { source: semver::Error },
+    #[snafu(display("The version exceeds the 128 character limit"))]
+    VersionTooLong,
```
NickLarsenNZ: This will then not be needed


```suggestion
```

rust/boil/src/cli/build.rs
```diff
@@ -129,16 +130,23 @@ pub struct BuildArguments {
     pub rest: Vec<String>,
 }
 
+// This is derived from the general rule where the length of the tag can be up to 128 chars
```
Techassi: note: Let's add a link to the upstream source this was derived from.

rust/boil/src/cmd/build.rs
```diff
@@ -44,13 +44,6 @@ pub enum Error {
 
 /// This is the `boil build` command handler function.
 pub fn run_command(args: Box<BuildArguments>, config: Config) -> Result<(), Error> {
-    // TODO (@Techassi): Parse Dockerfile instead to build the target graph
```
Techassi, line 47: note: That comment shouldn't be removed.

Techassi (commented):
Looks good to me overall. Just two small comments.

Can we also note down the reason why we want to relax this in the PR description for which exact use-case we need this?

rust/boil/src/cmd/build.rs
NickLarsenNZ: Ah ok, it was connected to that code (we should leave spaces when things aren't connected)

rust/boil/src/cli/build.rs
NickLarsenNZ:
```suggestion
// This is derived from the general rule where the length of the tag can be up to 128 chars
// See: https://github.com/opencontainers/distribution-spec/blob/main/spec.md
```

rust/boil/src/cmd/build.rs
```diff
@@ -44,13 +44,6 @@ pub enum Error {
 
 /// This is the `boil build` command handler function.
 pub fn run_command(args: Box<BuildArguments>, config: Config) -> Result<(), Error> {
```
NickLarsenNZ, line 46:
```suggestion
pub fn run_command(args: Box<BuildArguments>, config: Config) -> Result<(), Error> {
    // TODO (@Techassi): Parse Dockerfile instead to build the target graph
```

rust/boil/src/cmd/build.rs
NickLarsenNZ: See 54a9457

rust/boil/src/cli/build.rs
NickLarsenNZ: See 54a9457

Techassi (approved):

PR #1471, opened 2026-04-17
fix(vector): Use relative path to patchable checked out source

NickLarsenNZ:
Fixes this error:

```
find: '/src/': No such file or directory
```

Because for this (vector) build, WORKDIR is /stackable (so patchable checked out src/ is below that)

### DIFF
--- a/vector/copy_artifacts.sh
+++ b/vector/copy_artifacts.sh
@@ -6,5 +6,7 @@
 cp "$1" /app
 
 # And now try to find a BOM file named like the binary + _bin.cdx.xml and copy it over as well if it exists
+# Using `patchable with --images-repo-root=src` makes the path `/stackable/src/vector/patchable-work/worktree/0.52.0`
+# assuming WORKDIR was `/stackable`.
 base=$(basename "$1")
-find /src/ -type f -name "${base}_bin.cdx.xml" -exec cp {} /app \;
+find ./src/ -type f -name "${base}_bin.cdx.xml" -exec cp {} /app \;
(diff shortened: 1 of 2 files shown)


vector/copy_artifacts.sh
```diff
@@ -6,5 +6,7 @@
 cp "$1" /app
 
 # And now try to find a BOM file named like the binary + _bin.cdx.xml and copy it over as well if it exists
+# Using `patchable with --images-repo-root=src` makes the path `/stackable/src/vector/patchable-work/worktree/0.52.0`
+# assuming WORKDIR was `/stackable`.
 base=$(basename "$1")
-find /src/ -type f -name "${base}_bin.cdx.xml" -exec cp {} /app \;
+find ./src/ -type f -name "${base}_bin.cdx.xml" -exec cp {} /app \;
```
dervoeti:
```suggestion
find . -maxdepth 1 -type f -name "${base}_bin.cdx.xml" -exec cp {} /app \;
```

dervoeti (changes requested):
Build worked but the SBOM was not there (it's at the workspace level, not in `./src`). With the fix it is located here in the final image: `/usr/local/bin/vector_bin.cdx.xml`

vector/copy_artifacts.sh
NickLarsenNZ: I'm unsure if you tested it or not, but I now don't have time (due to other tasks) so I would go with what I have since it worked (it is gated by the exact match for `${base}_bin.cdx.xml`).

I'm happy to come back and fix it if you see an issue - just not sure if this was more stylistic or an actual problem you discovered.

dervoeti: I tested it, your version does not fail but the image does not contain the SBOM, I tested my version and the resulting image has the SBOM included.

NickLarsenNZ: Perfect, will pull your change in.

dervoeti (approved):

PR #1476, opened 2026-04-24
fix(vector): Use correct license

NickLarsenNZ:
I had inadvertently copied the wrong license into the vector image since building it from source.


sbernauer (approved):
O.o

PR #1492, opened 2026-05-20
ci: Publish to quay.io

Techassi:
Part of https://github.com/stackabletech/issues/issues/716, the repository layout is based on https://github.com/stackabletech/decisions/issues/83.

This PR adds quay.io as a publish target for all product container image builds. With that, we can natively push the images to quay.io instead of mirroring them from our Harbor instance.

Test CI run: https://github.com/stackabletech/docker-images/actions/runs/26161872014

### DIFF
--- a/.github/workflows/build_airflow.yaml
+++ b/.github/workflows/build_airflow.yaml
@@ -30,6 +30,7 @@ jobs:
     uses: ./.github/workflows/reusable_build_image.yaml
     secrets:
       harbor-robot-secret: ${{ secrets.HARBOR_ROBOT_SDP_GITHUB_ACTION_BUILD_SECRET }}
+      quay-robot-secret: ${{ secrets.QUAY_ROBOT_SDP_GITHUB_ACTION_BUILD_SECRET }}
       slack-token: ${{ secrets.SLACK_CONTAINER_IMAGE_TOKEN }}
     permissions:
       id-token: write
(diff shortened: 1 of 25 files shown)


dervoeti (commented):
Basically LGTM, but:
 - Did you take a look at `build_spark-connect-client.yaml`? It does not get pushed to quay but it's also in the `stackable` repository (not `sdp`) so maybe that's on purpose.

 - It seems like robot accounts on quay.io need to match a certain regex so the current name would fail? I just tested this with my own account:

Image

Techassi:
> Did you take a look at build_spark-connect-client.yaml? It does not get pushed to quay but it's also in the stackable repository (not sdp) so maybe that's on purpose.

Yeah, I think that is mostly an internal thing. I'm fine not pushing it to quay.io, but I'm also happy to push it - it is basically a decision we have to make.

> It seems like robot accounts on quay.io need to match a certain regex so the current name would fail? I just tested this with my own account:

The names should be fine, because I created the robot accounts using these names. Maybe it fails because they already exist?

image

dervoeti:
> Yeah, I think that is mostly an internal thing. I'm fine not pushing it to quay.io, but I'm also happy to push it - it is basically a decision we have to make.

Okay, we can decide later, won't block this PR for this.

> The names should be fine, because I created the robot accounts using these names. Maybe it fails because they already exist?

Ah! They get the `stackable+` prefix from quay. Then it works.

dervoeti (approved):

PR #1493, opened 2026-05-21
feat(testing-tools): Add grpcurl utility

NickLarsenNZ:
This is needed by Vector Aggregator now that the `/graphql` endpoint is gone

See https://github.com/stackabletech/decisions/issues/87

### DIFF
--- a/testing-tools/Dockerfile
+++ b/testing-tools/Dockerfile
@@ -1,6 +1,9 @@
 # syntax=docker/dockerfile:1.16.0@sha256:e2dd261f92e4b763d789984f6eab84be66ab4f5f08052316d8eb8f173593acf7
 # check=error=true
 
+ARG GRPCURL_VERSION
+FROM fullstorydev/grpcurl:v${GRPCURL_VERSION} AS grpcurl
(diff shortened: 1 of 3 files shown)


testing-tools/Dockerfile
```diff
@@ -1,6 +1,9 @@
 # syntax=docker/dockerfile:1.16.0@sha256:e2dd261f92e4b763d789984f6eab84be66ab4f5f08052316d8eb8f173593acf7
 # check=error=true
 
+ARG GRPCURL_VERSION
+FROM fullstorydev/grpcurl:v${GRPCURL_VERSION} AS grpcurl
```
NickLarsenNZ, line 5: Need to do something about this.
- Mirror
- Pullthrough cache

Techassi (approved):

testing-tools/Dockerfile
NickLarsenNZ: We decided to leave as-is and fix it if we have trouble.
It isn't a production image, otherwise we would spend the effort now.

PR #1496, opened 2026-05-21
ci: Add publish-to-quay toggle

Techassi:
This came up after merging #1492.

We consider this an "internal" image and as such don't see the need to publish it to quay. As such, this PR introduces a toggle which controls if images should be published to quay.io or not.

Test CI run: https://github.com/stackabletech/docker-images/actions/runs/26226133632


NickLarsenNZ (approved):
LGTM

PR #1503, opened 2026-05-21
feat(boil): Add support for remote layer caching

Techassi:
This PR adds support to enable remote caching for every layer (of all multi-stage) images.

Some first initial tests looked promising, but we have to keep an eye on push/upload performance. Also, we maybe want to experiment with the `min` mode.

The cached artifacts (for local testing) can be found here: `oci.stackable.tech/sdp-cache`.


rust/boil/README.md
```diff
@@ -31,3 +31,28 @@ boil image list
...
+
+boil offers to option to automatically pull from and push to a remote cache. This feature can be
+enabled by using the `--cache-registry` (and the optional `--cache-namespace`) argument:
+
+```shell
+# This will use `oci.example.org/<NAMESPACE>-cache/airflow` to store and retrieve cached layers
+boil build airflow --cache-registry oci.example.org
+
+# This will use `oci.example.org/foo/airflow` to store and retrieve cached layers
+boil build airflow --cache-registry oci.example.org --cache-namespace foo
```
NickLarsenNZ, line 96: Don't these need `-- --builder container` since the default doesn't support it? Or am I mistaken?

Techassi: It really depends on what people have selected as the default builder (driver). Those two examples are meant to show how to use the new args in _general_, which assumes the correct builder is used.

NickLarsenNZ (dismissed):
LGTM at a glance, but I haven't tried it.

Issue #1494, opened 2026-05-21
chore(hdfs): Update major/minor versions for 26.7.0


NickLarsenNZ:
Part of <https://github.com/stackabletech/docker-images/issues/1479>.

- Add: `3.4.3` (LTS), `3.5.0` (try it, maybe it becomes LTS?)
- Deprecate: `3.4.2`

## Update tasks

- [x] Update `boil-config.toml` to reflect the agreed upon versions in the spreadsheet (including the removal of old versions).
- [x] Upload new version (see `hadoop/upload_new_hadoop_version.sh`).
- [x] Update `boil-config.toml` to the latest supported version of JVM (base and devel).
- [x] Update other dependencies if applicable (eg: hdfs_utils, jmx_exporter, protobuf, etc).
- [x] Check other operators (getting_started / kuttl / supported-versions) for usage of the versions. Add the PR(s) to the list below.
- [x] Update the version in demos. Add the PR(s) to the list below.

## Related Pull Requests

> [!TIP]
> Delete any items that do not apply so that all applicable items can be checked.
> For example, if you add release notes to the documentation repository, you do not need the latter two criteria.

- https://github.com/stackabletech/docker-images/pull/1511
- https://github.com/stackabletech/hdfs-utils/pull/62
- https://github.com/stackabletech/hdfs-operator/pull/793
- https://github.com/stackabletech/hive-operator/pull/718
- https://github.com/stackabletech/trino-operator/pull/901
- https://github.com/stackabletech/nifi-operator/pull/947
- https://github.com/stackabletech/druid-operator/pull/831
- https://github.com/stackabletech/hbase-operator/pull/766
- https://github.com/stackabletech/spark-k8s-operator/pull/700
- https://github.com/stackabletech/demos/pull/410

## Acceptance

> [!TIP]
> This list should be completed by the assignee(s), once respective PRs have been merged. Once all items have been
> checked, the issue can be moved into _Development: Done_.

- [x] Can build image (either locally, or in CI)
- [ ] All kuttl tests pass (either locally, or in CI) => We will see
- [ ] ~Release notes added to documentation and linked as a PR above~ business as usual
- [ ] ~Release notes written in a comment below~ business as usual
- [ ] ~Applicable `release-note` label added to this issue~ business as usual

<details>
<summary>Testing instructions</summary>

```shell
# Where x.y.z is the new version added in this PR
boil build hadoop=x.y.z --strip-architecture --load

[...]

PR #1511, opened 2026-05-27
feat(hadoop): Add 3.5.0 and 3.4.3

sbernauer:
# Description

Part of https://github.com/stackabletech/docker-images/issues/1494
Needs https://github.com/stackabletech/hdfs-utils/pull/62

* Add Hadoop 3.4.3 and 3.5.0
* As all of the aws and azure lib collection is a clusterfuck, I remodeled it: The hdfs/hdfs image takes care of placing all cloud libs in `/stackable/hadoop-cloud-libraries/`, so that dependant images can easily pull them in. Side node: We now support 3 different Hadoop versions with 3 different naming conventions, argh.
* I tried to be a good citizen and bump Hadoop 3.4.2 -> 3.4.3 in downstream images. However, this caused compilation failures. As this PR is big enough I left it for the downstream version bumps.

> [!NOTE]

<details>

```shell
```

</details>


hadoop/hdfs-utils/stackable/patches/0.6.0/patchable.toml
```diff
@@ -0,0 +1,2 @@
+# FIXME: Use released commit
```
sbernauer: marker

.scripts/upload_new_boost_version.sh
NickLarsenNZ, line 1: Could we use patchable instead and avoid uploading sources to Nexus?

druid/boil-config.toml
NickLarsenNZ, line 1: I think Druid should be its own PR

.scripts/upload_new_boost_version.sh
NickLarsenNZ: As in, it becomes an image layer with a boil-config.toml and patches

NickLarsenNZ: It looks like you did, since there is "shared/boost". We just need to mirror the repo (if not done).

druid/boil-config.toml
NickLarsenNZ: I see all the other products were update in this PR too.
I guess it is fine, but maybe find the tracking issues to link this PR in them (did you do the whole task for all, or only hdfs related stuff?)

NickLarsenNZ: The reminder only have TODOs. How was druid different (why could it have the bump while the others couldn't)?

hadoop/hadoop/Dockerfile
```diff
@@ -126,3 +143,31 @@ rm -rf /stackable/.m2
 # Set correct groups; make sure only required artifacts for the final image are located in /stackable
 chmod -R g=u /stackable
 EOF
+
+RUN <<EOF
+NEW_VERSION=${PRODUCT_VERSION}-stackable${RELEASE_VERSION}
+
+# Place cloud libraries (for s3a:// and abfs://) in the /stackable/hadoop-cloud-libraries/ folder to
+# have one central place for dependant images to pull them from
+mkdir -p /stackable/hadoop-cloud-libraries/
```
NickLarsenNZ, line 152: Nice!

NickLarsenNZ (dismissed):
Looks fine. I didn't check in detail, but that should come out via testing.

I left a few comments for clarifications.

.scripts/upload_new_boost_version.sh
sbernauer: I copied the Dockerfile from @razvan and he documented it here: https://github.com/stackabletech/docker-images/blob/61741f635ee0bd769e59056e81c472c762462d83/shared/boost/Dockerfile#L10-L14

druid/boil-config.toml
sbernauer: I kept the ones where the compilation worked. All the ones where I was running into compilation errors which I couldn't solve in a sane timeframe I left as a TODO.
Such big changes in a patch level release suck... Any maven dependency resolution sucks, argh

hadoop/hadoop/Dockerfile
sbernauer: Thanks!

.scripts/upload_new_boost_version.sh
NickLarsenNZ: Ah ok, cool.

Is it something we want patchable to handle (submodules)?

We should also consider moving to https://codeberg.org/natkr/lappverk/

druid/boil-config.toml
NickLarsenNZ: Hmm yeah. Probably best to leave them to their tracking tasks. Otherwise it might get seen as done, and deeper testing gets skipped.

Efficient PRs don't necessarily mean an efficient process.

NickLarsenNZ (dismissed):

druid/boil-config.toml
sbernauer: I reverted all bumps in 01e6d40

.scripts/upload_new_boost_version.sh
sbernauer: I have to admit I personally am happy with the current state. We don't need to patch boost (and I don't expect we will), so I'm willing to save that work till we actually need it.

NickLarsenNZ: It's less about patching the source, and more about using tools to make it easier.
Avoiding another script to push stuff to nexus is good IMO. Same thing I did for statsd_exporter.

But we can leave it for later. I can sort that out when I think of it next.

NickLarsenNZ (approved):
LGTM

PR #1513, opened 2026-05-28
ci: Add publish toggle

Techassi:
This PR adds a toggle to the reusable action and the manual workflow dispatch to enable (default) or disable pushing and signing images _in general_.

It should be noted, that a `publish-to-quay` input is already available to control whether the images are pushed to quay.io.

> [!NOTE]
> Test CI runs:
> - https://github.com/stackabletech/docker-images/actions/runs/26561331798 (no skip)
> - https://github.com/stackabletech/docker-images/actions/runs/26563813501 (skip)
>
> There is unfortunately no way (before merging) for us to test other workflow triggers and ensure the defaulting of `input.skip-publish` works.


NickLarsenNZ (approved):
LGTM

Issue #1505, opened 2026-05-21
chore(trino): Update major/minor versions for 26.7.0


NickLarsenNZ:
Part of <https://github.com/stackabletech/docker-images/issues/1479>.

- Add: `481`
- Deprecate: `479`

### Trino

- [x] Update `boil-config.toml` to reflect the agreed upon versions in the spreadsheet (including the removal of old versions).
- [x] Update `boil-config.toml` to the latest supported version of JVM (base and devel).
- [x] Update other dependencies if applicable (eg: jmx_exporter, opa_authorizer, storage_connector, etc).
- [x] Check other operators (getting_started / kuttl / supported-versions) for usage of the versions. Add the PR(s) to the list below.
- [ ] Maybe update versions in `tests/templates/kuttl/opa-authorization/check-opa.py.j2` (if tests fail)
- [x] Update the version in demos. Add the PR(s) to the list below.
- [x] Update versions used in the documentation repository

### trino-cli

- [x] Update `trino-cli/boil-config.toml` to reflect the agreed upon versions in the spreadsheet (including the removal of old versions).
- [x] Upload new version (see `trino-cli/upload_new_trino_version.sh` scripts).
- [x] Update `boil-config.toml` to the latest supported version of JVM (base and devel).
- [x] Check other operators (getting_started / kuttl / supported-versions) for usage of the versions. Add the PR(s) to the list below.
- [x] Update the version in demos. Add the PR(s) to the list below.
- [x] Update versions used in the documentation repository

## Related Pull Requests

> [!TIP]
> Delete any items that do not apply so that all applicable items can be checked.
> For example, if you add release notes to the documentation repository, you do not need the latter two criteria.

- https://github.com/stackabletech/docker-images/pull/1518
- https://github.com/stackabletech/trino-operator/pull/900
- https://github.com/stackabletech/nifi-operator/pull/948
- https://github.com/stackabletech/demos/pull/412
- https://github.com/stackabletech/documentation/pull/859

## Acceptance

> [!TIP]
> This list should be completed by the assignee(s), once respective PRs have been merged. Once all items have been
> checked, the issue can be moved into _Development: Done_.

- [x] Can build image (either locally, or in CI)
- [ ] All kuttl tests pass (either locally, or in CI) -> At least trino-op did pass
- [ ] ~Release notes added to documentation and linked as a PR above~ nothing worth mentioning
- [ ] ~Release notes written in a comment below~ nothing worth mentioning
[...]

PR #1517, opened 2026-06-02
feat(trino): Add 481

Techassi:
Part of https://github.com/stackabletech/docker-images/issues/1505.

This PR adds Trino 481 and marks 479 as deprecated. The trino-storage-connector still uses Trino 477, as there is no upstream version for 481 (or 479).


Techassi:
Closing in favour of #1518.

PR #1527, opened 2026-06-12
feat(boil): Improve empty image version filter error

Techassi:
This now reports the image name and the filtered version(s) in the error. Previously, users would have to guess which image and version yielded an empty list.

### DIFF
--- a/deny.toml
+++ b/deny.toml
@@ -15,6 +15,37 @@ targets = [
 [advisories]
 yanked = "deny"
 ignore = [
+  # https://rustsec.org/advisories/RUSTSEC-2023-0071
+  # "rsa" crate: Marvin Attack: potential key recovery through timing sidechannel
+  #
+  # No patch is yet available, however work is underway to migrate to a fully constant-time implementation.
+  # So we need to accept this, as of SDP 26.3 we are "only" using the crate to create private +
+  # public key pairs used by webhooks, such as conversion or mutating webhooks.
+  #
+  # https://github.com/RustCrypto/RSA/issues/19 is the tracking issue
+  "RUSTSEC-2023-0071",
+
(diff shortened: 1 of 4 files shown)


deny.toml
sbernauer, line 1: Our source of truth is https://github.com/stackabletech/operator-templating/blob/main/template/deny.toml, can you pls update that first?

Techassi: Sure, but _this_ is not templated. So it should merge as is anyway.

sbernauer: Sure, technically we could merge this PR without problems, but I want to have a single place to look to see what rustsecs we currently silence

Techassi: operator-templating PR: https://github.com/stackabletech/operator-templating/pull/595

Techassi: The PR above is merged, so I assume you are now fine with this PR moving forwards?

sbernauer: There actually is quite a diff between upstream and this version. Can you pls pull the upstream version into this branch? (happy to do as well)

Techassi: Done in d860bc37.

sbernauer: Sorry, I might wasn't precise enough, I meant `cp ../operator-templating/template/deny.toml deny.toml
`

Techassi: Ah, I see!

Done in 3fd5187d.

NickLarsenNZ (approved):
LGTM

deny.toml
sbernauer: I'm sorry, but I don't understand what you are doing :see_no_evil: 
I'm talking about literally taking the template from operator-templating and pasting it here. I even gave the command :sweat_smile: 
Opened https://github.com/stackabletech/docker-images/pull/1529, pls have a look

PR #1545, opened 2026-06-19
docs: Clarify on the current state of druid opa authorizer

sbernauer:
# Description

As this is Druid I'm not forcing anyone to fix this, but I want to at least document it.
I would be even happier if I oversee something and this actually isn't a problem at all :)

> [!NOTE]

<details>

```shell
```

</details>


sbernauer:
@lfrancke convinced my this is ok to do (We compile against new versions and java compiler make sure signature stayed the same). If we fee like it we might document that in the authorizer's README.

PR #1555, opened 2026-06-30
feat(boil): Add support for floating tags

Techassi:

This PR adds support to add a floating tag to the generated Bakefile which ultimately is handed to docker to produce images with more than one tag.

These changes require changes in our actions. PR(s) for these will be raised soon.

### DIFF
--- a/rust/boil/Cargo.toml
+++ b/rust/boil/Cargo.toml
@@ -9,7 +9,9 @@ publish = false
 
 [dependencies]
 cap-std.workspace = true
-clap.workspace = true
+# We use the unstable-markdown feature to get proper markdown list rendering.
+# See tracking issue: https://github.com/clap-rs/clap/issues/5900
+clap = { workspace = true, features = ["unstable-markdown"] }
 clap_complete.workspace = true
 clap_complete_nushell.workspace = true
 git2.workspace = true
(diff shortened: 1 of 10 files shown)


rust/boil/src/utils.rs
```diff
@@ -1,7 +1,44 @@
...
+
+    fn floating(&self) -> String;
+}
+
+impl VersionExt for semver::Version {
+    fn is_floating(&self) -> bool {
+        self.major == 0
+            && self.minor == 0
+            && self.patch == 0
+            && (self.pre.starts_with("pr") || self.pre.as_str() == "dev")
```
NickLarsenNZ, line 21: If we want to keep this usable outside of stackable (like how we make the vendor_version a String) We could make this a configurable list of skipped tags (or tag prefixes/suffixes).

That isn't urgent, so it could be done in a future PR.

rust/boil/Cargo.toml
```diff
@@ -9,7 +9,9 @@ publish = false
 
 [dependencies]
 cap-std.workspace = true
-clap.workspace = true
+# We use the unstable-markdown feature to get proper markdown list rendering.
+# See tracking issue: https://github.com/clap-rs/clap/issues/5900
+clap = { workspace = true, features = ["unstable-markdown"] }
```
NickLarsenNZ, line 14: Image

NickLarsenNZ (approved):
LGTM

rust/boil/src/utils.rs
Techassi: Oh yeah, makes sense. Let's raise a separate PR for that.

Techassi: I've created an issue for this: https://github.com/stackabletech/docker-images/issues/1556

PR #1558, opened 2026-07-01
ci: Pin boil version to 0.2.2

Techassi:
This is done in order to safely publish and experiment with release candidates for boil. There will be at least one RC to introduce floating tag support and to migrate to a structured output for the `--write-image-manifest-uris` argument.


NickLarsenNZ (approved):
LGTM

PR #1566, opened 2026-07-06
fix: Collect provenance digests regardless of artifact layout

dervoeti:
# Description

`download-artifact` only nests each artifact into its own subdirectory when more than one matches the pattern. Single-version products upload a single provenance-digests-* artifact, which gets flattened directly into the download path, so the hardcoded `artifacts/*/digests.json` glob found nothing and `jq` failed. Use `find` so both the flat and per-artifact subdirectory layouts work, and guard against zero digest files.

> [!NOTE]

<details>

```shell
```

</details>

### DIFF
--- a/.github/workflows/reusable_build_image.yaml
+++ b/.github/workflows/reusable_build_image.yaml
@@ -268,6 +268,18 @@ jobs:
           set -euo pipefail
           IMAGE="${IMAGE_NAME:-$PRODUCT_NAME}"
 
+          # Collect every digests.json regardless of the directory layout that
+          # download-artifact produces.
+          readarray -t DIGEST_FILES < <(find artifacts -type f -name digests.json)
+
+          # No digests recorded means nothing to attach provenance to. Guard here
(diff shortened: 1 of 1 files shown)


NickLarsenNZ (approved):
LGTM

Techassi (approved):
LGTM in general, but just a question out of curiosity: Do we expect more than one file in some situations. If so, can you elaborate on these?

dervoeti:
> but just a question out of curiosity: Do we expect more than one file in some situations. If so, can you elaborate on these?

Yes, we have one `digests.json` per product version, so multiple files is even the more common case (and that one worked before, single version did not). The build matrix fans out over versions, each leg uploads its own `provenance-digests-<version>` artifact containing one `digests.json`.

  - Single-version products (krb5): 1 file.
  - Multi-version products: N files. e.g. kafka publishes 4 versions (3.9.1, 3.9.2, 4.1.1, 4.2.1) -> 4 `digests.json`

PR #1569, opened 2026-07-06
ci(boil): Checkout repo to fix gh command

Techassi:
The `gh` command needs a `.git` directory apparently. So now we checkout the repo when finishing the release.


NickLarsenNZ (approved):
LGTM

PR #1571, opened 2026-07-07
fix(boil): Improve VersionExt::is_floating impl

Techassi:
This commit contains two improvements to the `VersionExt::is_floating` function:

1. It relaxes the requirements for 0.0.0-dev versions to be able to include extra tag data, eg. 0.0.0-dev-pr123
2. It now uses a regular expression to check whether the prerelease string is considered floating. The previous solution was too limited to catch more complicated patterns.

Came up in https://github.com/stackabletech/actions/actions/runs/28854067368/job/85576153193?pr=116.

### DIFF
--- a/rust/boil/src/utils.rs
+++ b/rust/boil/src/utils.rs
@@ -1,8 +1,18 @@
-use std::{process::Command, str::FromStr};
+use std::{process::Command, str::FromStr, sync::LazyLock};
 
+use regex::Regex;
 use snafu::{ResultExt, Snafu};
 
 use crate::{cli::HostPort, core::platform::Architecture};
+
+/// A regular expression to check whether the prerelease string of a semantic version is considered
+/// floating.
+///
+/// Used in [`VersionExt::is_floating`].
+static SEMVER_PRERELEASE_FLOATING: LazyLock<Regex> = LazyLock::new(|| {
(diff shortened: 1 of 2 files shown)


sbernauer (dismissed):

rust/boil/src/utils.rs
```diff
@@ -1,9 +1,19 @@
...
+use regex::Regex;
 use snafu::{ResultExt, Snafu};
 
 use crate::{cli::HostPort, core::platform::Architecture};
 
+/// A regular expression to check whether the prerelease string of a semantic version is considered
+/// floating.
+///
+/// Used in [`VersionExt::is_floating`].
+static SEMVER_PRERELEASE_FLOATING: LazyLock<Regex> = LazyLock::new(|| {
```
NickLarsenNZ, line 12: not semver 😅

Techassi: I would say this is unrelated to the SemVer/CalVer discussion.

The `VersionExt` trait operates on `semver::Version` and there are some versions we consider floating. That's why I named it that way.

NickLarsenNZ: That's gonna leave a mark

NickLarsenNZ: But yeah, since it is on SemVer::Version, I have no argument. Only that we need to move away from SemVer for the SDP verison.

But yes, different discussion.

NickLarsenNZ (approved):
LGTM

PR #1584, opened 2026-07-20
chore(java-devel): Tweak Maven transport settings [skip ci]

Techassi:
> [!NOTE]
> This might get cherry-picked onto the `release-26.7` branch to potentially help with failing builds in CI.

This PR tweaks various Maven transport/resolver settings to be a little more network failure resistant in the future. These network issues seem to be known by the Java/Maven community and are related to Azure's network setup.

PR #1591, opened 2026-07-21
fix(java-devel): Apply Maven retry settings under connector.* keys used by Maven 3.9

lfrancke:
## Description

In [#1584] @Techassi added HTTP retry settings to `java-devel/stackable/settings.xml`, but only under the `aether.transport.http.*` keys. That prefix is the maven-resolver 2.x / Maven 4 spelling. We build with Maven 3.9.16 (maven-resolver 1.9.27), which only reads `aether.connector.http.*. So those retry settings currently have no effect. I diagnosed that years ago and we even have a comment in the file about it.

This adds the `connector.*` spelling alongside the existing `transport.*` one, so the retries actually apply today, mirroring how `connectionMaxTtl` is already dual-spelled in the same file. The `transport.*` keys are kept for the eventual Maven 4 upgrade.

[#1584]: https://github.com/stackabletech/docker-images/pull/1584

### DIFF
--- a/java-devel/stackable/settings.xml
+++ b/java-devel/stackable/settings.xml
@@ -59,21 +59,33 @@
         <aether.transport.http.connectionMaxTtl>25</aether.transport.http.connectionMaxTtl>
 
         <!--
+        As with connectionMaxTtl above, the retry keys were renamed between maven-resolver 1.x
+        (aether.connector.http.*) and 2.x (aether.transport.http.*).
+        At the time of writing we were building with a version of Maven that requires the old ("connector") properties.
+        The "transport" keys are here for Maven 4.
+        Verified against ConfigurationProperties.java @ maven-resolver-1.9.27.
+        It does not hurt to keep both sets in here.
+        -->
+
+        <!--
         By default, requests are retried 3 times. We increase the count slightly to hopefully get less failing
         builds in CI.
         -->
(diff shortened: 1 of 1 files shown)


CHANGELOG.md
```diff
@@ -37,6 +31,7 @@ All notable changes to this project will be documented in this file.
 
 ### Changed
 
+- java-devel: Apply the Maven HTTP retry settings under the `aether.connector.*` keys read by Maven 3.9 so they actually take effect; the previously added `aether.transport.*` keys are Maven 4 only ([#1591]).
```
Techassi: This change needs to move, because #1590 was merged onto main.

Techassi: Also, I didn't mention the initial PR in the changelog. So should we either remove the entry for this PR, or retroactively add an entry for my change? That would also need to be cherry-picked onto the release branch.

Techassi (dismissed):
Lgtm, just a minor comment about the changelog.

lfrancke:
Thanks! Rebased onto `main` (past #1590's release changelog update) and **removed the changelog entry** — consistent with #1584/#1585, which didn't have one either, and it avoids a release-branch cherry-pick. The PR is now just the `settings.xml` fix.

Techassi (approved):

Techassi:
I would hold off on merging this for a little while until the 26.7.0 images are out of the door.

PR #1592, opened 2026-07-21
Add a script to watch builts for a release tag

lfrancke:
# Description

This is just a simple bash script to watch the output of `gh run list --branch <tag>` and restart all builds that have failed. Helpful during release when things are failing...


NickLarsenNZ (approved):
Depending on how integrated we want boil to be, this functionality could move there.

But as it is, this is already useful (and opt-in).

PR #1598, opened 2026-08-13
feat(boil)!: Output structured data in boil-target-tags file

Techassi:
> [!CAUTION]
> This requires changes in the `build-product-image` action.

Our actions/workflows need to know which tag boil considers the canonical tag in order to point to the correct fully-qualified image tags in the image index manifest.

As such, the `boil-target-tags` file now contains structured (JSON) data in the following form:

```json
{
  "krb5": {
    "canonical": "localhost/sdp/krb5:1.21.1-stackable26.7.0-amd64",
    "others": [
      "localhost/sdp/krb5:1.21.1-stackable26.7-amd64"
    ]
  },
  "testing-tools": {
    "canonical": "localhost/sdp/testing-tools:0.3.0-stackable26.7.0-amd64",
    "others": [
      "localhost/sdp/testing-tools:0.3.0-stackable26.7-amd64"
    ]
  }
}
```

<details>
<summary>Command which produced the above output</summary>

```shell
boil build krb5=1.21.1 testing-tools=0.3.0 --image-version 26.7.0 --use-localhost-registry --write-image-manifest-uris --floating-tag --dry
```

</details>

### Breaking changes

- The default name of the output file produced by the `--write-image-manifest-uris` argument is now `boil-target-tags.json` instead of just `boil-target-tags`.
- The `boil-target-tags.json` now contains JSON instead of plain text.


.gitignore
```diff
@@ -5,3 +5,6 @@ target/
 
 # Patchable working files
 patchable-work/
+
+# Generated by boil
+boil-target-tags.json
```
NickLarsenNZ, line 10: Just curious, why is the file ignored? Isn't it generated an CI and not committed?

NickLarsenNZ: My bad. For some reason I thought this was .gitattributes

rust/boil/src/cmd/build.rs
```diff
@@ -51,8 +54,9 @@ pub fn run_command(args: Box<BuildArguments>, config: Config) -> Result<(), Erro
 
     // Write the image manifest URIs to file if requested
     if let Some(path) = args.write_image_manifest_uris {
-        std::fs::write(path, image_manifest_uris.join("\n"))
-            .context(WriteImageManifestUrisFileSnafu)?;
+        let file = std::fs::File::create(path).context(CreateImageManifestUrisFileSnafu)?;
+        serde_json::to_writer(file, &image_manifest_uris)
+            .context(SerializeImageManifestUrisFileSnafu)?;
```
NickLarsenNZ, line 59: Would we not guarantee it can serialize? Or is there something that could make it fail at runtime?


```suggestion
            .expect("image_manifest_uris is always serializable");
```

NickLarsenNZ (approved):
LGTM

rust/boil/src/cmd/build.rs
NickLarsenNZ: Have approved, will le you decide if you want do the above.

Techassi: I've went through various scenarios in my head and I _think_ using `expect` here _should_ be fine. As long as the serialized type doesn't change.

But honestly, we should stay on the safe side and properly handle any error (even if it is basically impossible to hit - but never disregard the impossible :smile:)

.gitignore
Techassi: I added it to the `.gitignore` because it can also be generated locally during testing/development.

PR #1600, opened 2026-08-17
feat: Add SBOMs for frontend (npm) dependencies

dervoeti:
# Description

Airflow, Druid, NiFi and Superset all bundle a frontend (npm) build, but their SBOM tooling (`cyclonedx-py`, CycloneDX Maven plugin) only covers the backend language. This adds a separate `cdxgen`-generated CycloneDX SBOM for the npm dependencies of each product's frontend, alongside the existing backend SBOM.

`cdxgen` 13.x requires Node >= 24, which is newer than the Node versions these products pin for their own frontend builds. Rather than bumping those, each affected Dockerfile installs a second, independent Node under `/opt/node-cdxgen` used only to run `cdxgen`, with that directory prepended to `PATH` for the `cdxgen` invocation.

As part of this, opensearch-dashboards (which already generated a frontend SBOM) is migrated from `cdxgen` 12.0.0 (`@cyclonedx/cdxgen`) to 13.0.1 (`@cdxgen/cdxgen`, the package was renamed for
13.x), and pins `--spec-version 1.6` explicitly, since cdxgen's default spec version changes between releases and 1.6 is
the lowest version cdxgen 13 accepts as a generation target.

We still have some more uncovered JS code in our products (e.g. Trino UI), that will be a separate PR though since this is likely more complicated.


druid/Dockerfile
```diff
@@ -33,6 +36,15 @@ microdnf install python-pyyaml
 
 microdnf clean all
 rm -rf /var/cache/yum
+
+# Install Node and cdxgen to create the SBOM of the web console (npm) dependencies.
+# The CycloneDX Maven plugin only covers the Java dependencies, and the Node version that
+# the frontend-maven-plugin downloads for the actual build is not necessarily compatible
+# with cdxgen, which requires Node 20, 22 or 24.
+ARCH="${TARGETARCH/amd64/x64}"
```
Techassi, line 47: Is it really `x64`? Not `x86_64`? If so, very weird decision man :O

druid/Dockerfile
```diff
@@ -33,6 +36,15 @@ microdnf install python-pyyaml
...
+curl "https://repo.stackable.tech/repository/packages/node/node-v${NODEJS_VERSION}-linux-${ARCH}.tar.xz" | \
```
Techassi: I would add something like `-fsSL` here. Or do we have access to our shared `curlrc` file here?. If so, maybe add a comment here stating that we do.

Techassi: Also, is this the standard way we install Node in other images as well? _Just asking to ensure consistency._

airflow/Dockerfile
```diff
@@ -142,6 +147,21 @@ if [ -d "./airflow-core" ]; then
...
+  # --no-babel is required in combination with --required-only: the babel based usage
+  #   analysis marks every package as optional unless it is imported directly, which would
+  #   also drop all transitive runtime dependencies.
+  # --project-version is passed because the frontends declare a placeholder version
+  #   upstream.
+  cdxgen \
+    --type js \
+    --required-only \
+    --no-babel \
+    --spec-version 1.5 \
```
Techassi: We might want this to be a global `build-argument` instead of hard-coding it in every single Dockerfile.

Techassi (commented):
Generally looks good, just a few comments.

druid/Dockerfile
NickLarsenNZ: Wait, what? Shouldn't this be x86_64? Or did they invent their own way?

dervoeti: Yes, unfortunately. x64 is what Node itself calls the architecture in its release tarballs.

druid/Dockerfile
dervoeti: Yes we are using our curlrc file here. I can add a comment.

> this the standard way we install Node in other images as well?

I copied it from https://github.com/stackabletech/docker-images/blob/main/opensearch-dashboards/Dockerfile

druid/Dockerfile
Techassi: Oh wow... That... is... a decision you can make. I guess :/

Techassi: Now I'm curious: We don't need a mapping for aarch64/arm64? If so, that makes the usage of `x64` even weirder.

airflow/Dockerfile
Techassi: Addressed in ae41590.

druid/Dockerfile
Techassi: The curl comment was addressed in ae41590.

druid/Dockerfile
dervoeti: It's arm64 and x64 :man_shrugging: 
https://nodejs.org/en/download/archive/v26.7.0

Techassi: Yikes!

airflow/Dockerfile
```diff
@@ -142,6 +161,21 @@ if [ -d "./airflow-core" ]; then
...
+  #   also drop all transitive runtime dependencies.
+  # --project-version is passed because the frontends declare a placeholder version
+  #   upstream.
+  PATH="/opt/node-cdxgen/bin:$PATH" cdxgen \
+    --type js \
+    --required-only \
+    --no-babel \
+    --spec-version ${CDXGEN_SPEC_VERSION} \
+    --project-version ${PRODUCT_VERSION} \
+    --output /stackable/app/airflow-ui-${PRODUCT_VERSION}.cdx.json
```
Techassi: We should quote those values.


```suggestion
    --spec-version "${CDXGEN_SPEC_VERSION}" \
    --project-version "${PRODUCT_VERSION}" \
    --output "/stackable/app/airflow-ui-${PRODUCT_VERSION}.cdx.json"
```

Techassi: Basically applies to all other files as well.

airflow/Dockerfile
```diff
@@ -159,6 +193,18 @@ else
...
+  # NOTE: A Yarn 1 lockfile does not record whether a package is a dev dependency, so
+  # cdxgen marks everything as required and --required-only has no effect here. The SBOM
+  # for Airflow 2.x therefore also contains the frontend build tooling.
+  PATH="/opt/node-cdxgen/bin:$PATH" cdxgen \
+    --type js \
+    --required-only \
+    --no-babel \
+    --spec-version ${CDXGEN_SPEC_VERSION} \
+    --project-version ${PRODUCT_VERSION} \
+    --output /stackable/app/airflow-www-${PRODUCT_VERSION}.cdx.json
```
Techassi:
```suggestion
    --spec-version "${CDXGEN_SPEC_VERSION}" \
    --project-version "${PRODUCT_VERSION}" \
    --output "/stackable/app/airflow-www-${PRODUCT_VERSION}.cdx.json"
```

opensearch-dashboards/Dockerfile
```diff
@@ -184,7 +195,9 @@ for plugin in $(ls); do
 done
 EOF
 
-RUN cdxgen --target yarn --json-pretty
+# --spec-version is passed explicitly because cdxgen changes its default between
+# releases. The value is a global build argument, see boil.toml.
+RUN PATH="/opt/node-cdxgen/bin:$PATH" cdxgen --target yarn --json-pretty --spec-version ${CDXGEN_SPEC_VERSION}
```
Techassi:
```suggestion
RUN PATH="/opt/node-cdxgen/bin:$PATH" cdxgen --target yarn --json-pretty --spec-version "${CDXGEN_SPEC_VERSION}"
```

Techassi (changes requested):

CHANGELOG.md
```diff
@@ -4,6 +4,14 @@ All notable changes to this project will be documented in this file.
 
 ## [Unreleased]
 
+### Added
+
+- airflow, superset, druid, nifi: Add SBOMs for the frontend (npm) dependencies ([#1600]).
+
+### Changed
+
+- opensearch-dashboards: Bump cdxgen to 13.0.1 and pin the CycloneDX spec version to 1.6 ([#1600]).
```
Techassi, line 13: I guess we technically do this for the others (airflow, superset, druid and nifi) as well, right?

Techassi (commented):
Approved, just one last question.

Techassi (approved):
Well approving would actually help.

CHANGELOG.md
dervoeti: Yes, everything uses cdxgen 13.0.1 and CycloneDX 1.6. For opensearch-dashboards this is a bump, for all others it's a new addition (the SBOMs there were missing before).

PR #1601, opened 2026-08-18
fix(boil): Include prerelease data in floating tag

Techassi:
The prerelease data should be included in the floating tag `boil` produces, eg:

- `1.2.3-pr321` -> `1.2-pr321` (without the fix, it produces `1.2`)
- `1.2.3` -> `1.2`


NickLarsenNZ (approved):
LGTM

PR #1603, opened 2026-08-18
feat(boil): Support for per-image build arguments

Techassi:
I noticed in #1600 that we need to copy+paste many build arguments (with the same value) across versions of the same image in the `boil-config` files. In order to make this easier, it is now possible to specify a shared list of build arguments which apply to all versions of a single image.

```toml
# Set it here once instead of having to set it for every version individually below.
[build-arguments]
golang-version = "1.26.0"

# Version 1.12.3
[versions."1.12.3".local-images]
stackable-devel = "1.0.0"
vector = "0.55.0"

# Version 1.16.2
[versions."1.16.2".local-images]
stackable-devel = "1.0.0"
vector = "0.55.0"
```

The arguments can be overwritten on a per-version basis if needed.

### DIFF
--- a/rust/boil/src/core/bakefile.rs
+++ b/rust/boil/src/core/bakefile.rs
@@ -311,10 +311,12 @@ impl Bakefile {
         let date_time = Self::now()?;
 
         // Load build arguments from a file if the user requested it
-        let mut user_container_build_args = cli_args.build_arguments.clone();
+        let mut user_container_build_args: docker::BuildArguments =
+            docker::build_args_vec_to_btree_map(cli_args.build_arguments.clone());
+
         if let Some(path) = &cli_args.build_arguments_file {
             let build_arguments_from_file =
-                docker::BuildArguments::from_file(path).context(ParseBuildArgumentsSnafu)?;
+                docker::build_args_from_file(path).context(ParseBuildArgumentsSnafu)?;
             user_container_build_args.extend(build_arguments_from_file);
         }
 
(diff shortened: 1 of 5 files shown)


rust/boil/src/core/bakefile.rs
```diff
@@ -385,7 +385,8 @@ impl Bakefile {
 
                 // TODO (@Techassi): Clean this up
                 // TODO (@Techassi): Move the arg formatting into functions
-                let mut build_arguments = docker::BuildArguments::new();
+                // Start of with the shared (across all versions of the same image) build arguments.
```
dervoeti:
```suggestion
                // Start off with the shared (across all versions of the same image) build arguments.
```

dervoeti (changes requested):
> This feature currently doesn't allow overwriting the value on a per version level.

We need this to be in this PR I'd say because currently overwriting makes the build fail:
```
head -n13 opa/boil-config.toml
[metadata.registries]
"oci.stackable.tech" = { namespace = "sdp" }

[build-arguments]
golang-version = "9.9.9"

# Version 1.12.3
[versions."1.12.3".local-images]
stackable-devel = "1.0.0"
vector = "0.55.0"

[versions."1.12.3".build-arguments]
golang-version = "1.26.0"
```

```
cargo boil build opa=1.12.3
    Finished `dev` profile [unoptimized + debuginfo] target(s) in 0.11s
     Running `target/debug/boil build opa=1.12.3`
[+] Building 0.0s (1/1) FINISHED                                                                                                                                                     docker:default
 => [internal] load local bake definitions                                                                                                                                                     0.0s
 => => reading from stdin 4.64kB                                                                                                                                                               0.0s
-.json:1
--------------------
   1 | >>>
--------------------
ERROR: -.json:1,987-1003: Duplicate object attribute; An attribute named "GOLANG_VERSION" was already defined at -.json:1,961-977.
Error: failed to run build command
```

Which is probably even better than silently ignoring it, but I would still include that functionality in this PR already.

Cool feature though, makes the boil configs more readable :slightly_smiling_face:

Techassi:
> We need this to be in this PR I'd say because currently overwriting makes the build fail

Your wish has been granted in 1b0ce2d8 :)

dervoeti (dismissed):
Change itself LGTM, two optional things:
 - Update README.md
 - The build arguments in the global config (src/config.rs) should probably get
`#[serde(default, deserialize_with = "docker::deserialize_args")]` as well

Techassi:
Both very valid points. On it!

**EDIT:** Both addressed in a5ccbd23 and 6e856bc5.

dervoeti (approved):
ship it

PR #1605, opened 2026-08-19
feat(boil): Add circular local image dependency detection

Techassi:
This solves a long standing TODO in the target assembly.

Triggering a stack overflow due to the missing circular dependency detection was as simple as:

```toml
# in test1/boil-config.toml
[versions."1.0.0".local-images]
test2 = "1.0.0"

# in test2/boil-config.toml
[versions."1.0.0".local-images]
test1 = "1.0.0"
```

### DIFF
--- a/rust/boil/src/core/bakefile.rs
+++ b/rust/boil/src/core/bakefile.rs
@@ -86,6 +86,12 @@ pub enum TargetsError {
         versions: Vec<String>,
         image_name: String,
     },
+
+    #[snafu(display(
+        "failed to resolve local image chain. Circular dependency found: {}",
+        chain.iter().map(|(n, v)| format!("{n}={v}")).collect::<Vec<_>>().join(" -> ")
+    ))]
+    CircularDependency { chain: Vec<(String, String)> },
 }
 
 #[derive(Debug, Default)]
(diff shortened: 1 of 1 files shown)


rust/boil/src/core/bakefile.rs
```diff
@@ -247,8 +281,12 @@ impl Targets {
                     );
 
                     // Wowzers, recursion!
-                    self.insert_targets(image_name.clone(), image_config, options, false)?;
+                    self.insert_targets(image_name.clone(), image_config, options, false, chain)?;
                 }
+
+                // Remove the last target node once we are done with the current image name + image
+                // version combination.
+                chain.pop();
```
NickLarsenNZ: Why do we pop here?
I couldn't tell from the function signature. Maybe needs a doc.

NickLarsenNZ (dismissed):
LGTM

rust/boil/src/core/bakefile.rs
Techassi: I hopefully improved the comment in 1cbb02fe.

NickLarsenNZ (approved):
LGTM

rust/boil/src/core/bakefile.rs
NickLarsenNZ: Perfect

PR #1606, opened 2026-08-21
fix(boil): Only include prerelease data if present

Techassi:
Déjà vu is hitting hard.

Previously, this would result in `1.2-` as the floating tag if `1.2.3` is used as the input.


rust/boil/src/utils.rs
```diff
@@ -168,4 +169,22 @@ mod tests {
             semver::Version::from_str(input).expect("input must be valid semantic version");
         assert!(!version.is_floating());
     }
+
+    #[rstest]
```
NickLarsenNZ, line 173: praise

NickLarsenNZ (approved):

PR #1618, opened 2026-08-24
ci: Replace pre-commit with prek

Techassi:
This switch has been done in various other places. So it makes sense to also do it here. Additionally, we make use of prek workspaces to optimize which hooks we run based on changes.


NickLarsenNZ (approved):
LGTM

Ping me when merged, and I'll update the required check.

Existing PRs will need to merge main.

PR #1623, opened 2026-08-27
fix(superset): Remove NodeJS from final image

dervoeti:
# Description

The Superset image ships a full Node and npm installation that is only needed to build the frontend.

The reason is the download location: `nvm.sh` was written to `/stackable/nvm.sh`, and nvm derives its installation directory from the location of that script when `NVM_DIR` is unset. Node therefore ends up in `/stackable/versions`, and the final image copies `/stackable/` from the builder as a whole.

Setting `NVM_DIR=/opt/nvm` and downloading `nvm.sh` there keeps the whole nvm tree outside the copied directory. This mirrors what the cdxgen Node installation in `/opt/node-cdxgen` already does.

Besides the smaller image, this also removes a Node runtime that is never executed but is picked up by the vulnerability scans.

Other images were checked and are not affected: Airflow, NiFi, Kafka, Trino and OPA ship no Node at all, and OpenSearch Dashboards needs it at runtime.

> [!NOTE]

<details>

```shell
```

</details>


Techassi (approved):

lfrancke:
Very nice!

PR #1626, opened 2026-08-31
chore(java-devel): Disable Maven connection pooling

Techassi:
This hopefully reduces the number of network issues between the GitHub Action runners (Azure) and our Maven repo (IONOS). This is pretty much one of the last stop we can pull before we would need to investigate moving our Maven repo off of IONOS which hopefully has better networking to Azure or try out Pulp afterall and see if it improves the situation.

### DIFF
--- a/java-devel/stackable/settings.xml
+++ b/java-devel/stackable/settings.xml
@@ -89,9 +89,10 @@
         <aether.transport.http.retryHandler.intervalMax>60000</aether.transport.http.retryHandler.intervalMax>
 
         <!--
-        This is an option we can experiment with in the future. Settings this to false will result in no connection
-        pooling, which should help to avoid stale/long-lasting connections which can become inactive and be killed by
-        the Azure networking.
+        Disable connection pooling entirely (which makes the connectionMaxTtl settings redundant for now) to rule out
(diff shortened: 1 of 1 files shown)


java-devel/stackable/settings.xml
```diff
@@ -93,7 +93,8 @@
         pooling, which should help to avoid stale/long-lasting connections which can become inactive and be killed by
         the Azure networking.
         -->
```
lfrancke, line 96: Happy to experiment with this but then we need to update this comment

Something like:
Disable connection pooling entirely (which makes the connectionMaxTtl settings redundant for now) to rule out failures we're seeing where connections seem to be dropped without us noticing.
The downside is a fresh connection for everything we download.
We can disable this again should we find it doesn't help.

lfrancke (changes requested):

java-devel/stackable/settings.xml
Techassi: Yes, true. Makes sense. I will update it!

Techassi: Updated in dd317ab4.

lfrancke (approved):

PR #1627, opened 2026-09-02
ci: Enable floating tags

Techassi:

> [!NOTE]
> These changes were tested as best as they could here:
>
> - https://github.com/stackabletech/docker-images/actions/runs/33633919558 (no push/signing)
> - https://github.com/stackabletech/docker-images/actions/runs/33636217543
> - https://github.com/stackabletech/docker-images/actions/runs/33636408316
>
> It should be noted, that 0.0.0-dev is already considered floating and as such **no** extra tag is produced. Testing with a version which will produce a second tag will be done later.
>
> - https://github.com/stackabletech/docker-images/actions/runs/33638912230 (1.2.3 hard-coded, with push/sign to Harbor, artifacts were deleted after testing)

This PR wires up floating tag support for all product images. We currently still default to false to make sure everything continues to work as expected (and without changing behaviour).

### DIFF
--- a/.github/workflows/build_airflow.yaml
+++ b/.github/workflows/build_airflow.yaml
@@ -6,8 +6,13 @@ run-name: |
 on:
   workflow_dispatch:
     inputs:
-      skip-publish:
-        description: Skip publishing and signing images
+      publish:
+        description: Publish and sign images
+        type: boolean
+        required: true
+        default: true
+      floating-tag:
+        description: Produce floating tag
         type: boolean
         required: true
         default: false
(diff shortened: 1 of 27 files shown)


dervoeti (commented):
Regarding the `defaults` job, would something like this work maybe?
```
      with:
        product-name: krb5
        sdp-version: ${{ case(github.ref_type == 'tag', github.ref_name, '0.0.0-dev') }}
        registry-namespace: sdp
        # Keep these defaults in sync with the workflow_dispatch input defaults above.
        publish: ${{ github.event_name != 'workflow_dispatch' || inputs.publish }}
        floating-tag: ${{ github.event_name == 'workflow_dispatch' && inputs.floating-tag }}
```

Would shorten the whole code quite a bit.

dervoeti (approved):
LGTM, just one optional suggestion to make the code more compact.

Techassi:
> Regarding the `defaults` job, would something like this work maybe?

I honestly don't know, because empty values behave very weirdly in GHA. They only sometimes become a faulty value. This definitely would need a round of testing.

I'm happy to explore this option, but would prefer to do so in a separate PR.

PR #1630, opened 2026-09-07
fix: Remove non-shipped dependencies from the SBOMs

dervoeti:
# Description

Our SBOMs report dependencies that are never shipped, like test frameworks, build tooling and integrations that are not compiled in. They show up as false positives in the vulnerability scans.

Vector: builds with an explicit, reduced feature set, but `cargo cyclonedx` resolved the default features and reported every integration (AWS, Azure, etc.).

Rust base images: now pass `--no-build-deps`, which drops the build-time dependencies.

Kafka: extended the list of skipped projects

OpenSearch (3.1.0 and 3.4.0): additionally skips the test framework, the test fixtures, the QA projects and the benchmarks. OpenSearch 3.6.0 was already fine since there an allow-list instead of a deny-list is used.
The Prometheus exporter 3.6.0.0 gets `includeConfigs = ["runtimeClasspath"]` on its `cyclonedxDirectBom` task, the `cyclonedxBom` extension only configures the aggregation there. 3.1.0.0 and 3.4.0.0 use plugin version 2.4.1, where the extension still applies to the project itself, so they are already correct.

OpenSearch Dashboards and the HBase web UI pass `--required-only` to cdxgen, which leaves out devDependencies and optional dependencies.

Airflow and Superset got two fixes:

  * `cyclonedx-bom` was installed into the product environment, so it was part of its own SBOM, and uninstalling it afterwards left its dependencies behind in the image. It is now installed into a throw-away venv that is removed after the SBOM is written.
  * The Airflow and Superset packages are installed from a locally built wheel, so `cyclonedx-py` records the file URL of that wheel instead of a purl. Without a purl they appear twice in the image SBOM (no deduplication possible). The purl is added in post-processing with `jq`.

> [!NOTE]

<details>

```shell
```

</details>

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -22,6 +22,14 @@ All notable changes to this project will be documented in this file.
 - superset: Fix the broken builds by excluding the `cypress-base` end-to-end test project from the frontend SBOM ([#1616]).
 - superset: Fix the broken 4.1.4 build by also excluding `packages/superset-ui-switchboard` from the frontend SBOM ([#1620]).
 - superset: Install nvm into `/opt/nvm` so that Node and npm, which are only needed to build the frontend, are no longer shipped in the final image (about 161 MB) ([#1623]).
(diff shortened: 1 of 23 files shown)


StefanFl (approved):
lgtm

PR #1635, opened 2026-09-10
fix: Correct the SBOM scoping follow-ups from #1630

dervoeti:
# Description

Follow-up to #1630. While verifying the resulting image SBOMs, a few of the changes there turned out
to be incomplete, ineffective or to have removed too much.

OpenSearch Dashboards: `--target yarn` is not a cdxgen option, cdxgen silently ignored it and auto-detected every ecosystem in the source tree, so the GitHub Actions workflows and the Gradle and Python files of the build tooling ended up in the SBOM. 

Kafka `4.1.1` and `4.2.1`: `includeConfigs` now also contains `releaseOnly`, a few components were shipped without being in the SBOM.

Kafka (all versions): additionally skips `generator`, the integration test projects and `connect:test-plugins`.

OpenSearch `3.1.0` and `3.4.0`: additionally skip `build-tools` and the remaining unshipped projects (`sandbox`, `docs`, `doc-tools`, `rest-api-spec`, the client benchmarks and `client:sniffer`).

Airflow and Superset: the purl fix from #1630 only covered the Airflow and Superset packages themselves. It now applies to every package without a purl (`opa-auth-manager`, `opa-authorizer`, `pip`), and the names in all pypi purls are normalized as defined in the Python packaging specification. cyclonedx-py keeps the name as written in the metadata (`pkg:pypi/zope.event`) while Syft normalizes it (`pkg:pypi/zope-event`), which was another source of duplicates in the image SBOM.

> [!NOTE]

<details>

```shell
```

</details>

### DIFF
--- a/airflow/Dockerfile
+++ b/airflow/Dockerfile
@@ -234,15 +234,25 @@ rm -rf /tmp/sbom-venv
 
 # Post-process the SBOM:
 #   * Break circular dependencies by removing the apache-airflow dependency from the providers.
-#   * Add the missing purl to the Airflow packages. They are installed from a locally built
-#     wheel, so cyclonedx-py records the file URL of that wheel instead of a purl.
-jq '.dependencies |= map(if .ref | test("^apache-airflow-providers-") then
+#   * Add the missing purls. Packages that are installed from a local wheel or a direct URL
+#     (apache-airflow, opa-auth-manager, pip) get no purl from cyclonedx-py, because it records
+#     the file URL instead. Without a purl they show up twice in the image SBOM.
+#   * Normalize the names in the purls as defined in the Python packaging specification.
+#     cyclonedx-py keeps the name as it is written in the metadata (`pkg:pypi/zope.event`),
+#     while Syft normalizes it (`pkg:pypi/zope-event`), which again results in duplicates.
+jq 'def normalize_name: ascii_downcase | gsub("[-_.]+"; "-");
+.dependencies |= map(if .ref | test("^apache-airflow-providers-") then
     .dependsOn |= map(select(. != "apache-airflow=='${PRODUCT_VERSION}'"))
 else
     .
 end)
-| .components |= map(if (.purl | not) and (.name | test("^apache[-_]airflow")) then
-    .purl = "pkg:pypi/" + (.name | ascii_downcase | gsub("[-_.]+"; "-")) + "@" + .version
+| .components |= map(if (.purl | not) and (.name != null) and (.version != null) then
+    .purl = "pkg:pypi/" + (.name | normalize_name) + "@" + .version
+else
+    .
+end
+| if (.purl // "") | startswith("pkg:pypi/") then
+    .purl |= sub("^pkg:pypi/(?<name>[^@]+)"; "pkg:pypi/" + (.name | normalize_name))
 else
     .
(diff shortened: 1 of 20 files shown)


airflow/Dockerfile
```diff
@@ -234,15 +234,25 @@ rm -rf /tmp/sbom-venv
...
 end)
-| .components |= map(if (.purl | not) and (.name | test("^apache[-_]airflow")) then
-    .purl = "pkg:pypi/" + (.name | ascii_downcase | gsub("[-_.]+"; "-")) + "@" + .version
+| .components |= map(if (.purl | not) and (.name != null) and (.version != null) then
+    .purl = "pkg:pypi/" + (.name | normalize_name) + "@" + .version
+else
+    .
+end
+| if (.purl // "") | startswith("pkg:pypi/") then
+    .purl |= sub("^pkg:pypi/(?<name>[^@]+)"; "pkg:pypi/" + (.name | normalize_name))
```
NickLarsenNZ, line 255: **No action needed**

This is a little hard to read, but I tried formatting it myself and it isn't much better.

In any case, thought I'd let you know that can can provide args to jq and use them in expressions so that you don't need to close the single quote just to use a shell variable).

There are two ways:

Via binding:

```suggestion
jq --arg product_version "$PRODUCT_VERSION" 'def normalize_name: ascii_downcase | gsub("[-_.]+"; "-");
.dependencies |= map(if .ref | test("^apache-airflow-providers-") then
   .dependsOn |= map(select(. != "apache-airflow==" + $product_version))
else
   .
end)
| .components |= map(if (.purl | not) and (.name != null) and (.version != null) then
    .purl = "pkg:pypi/" + (.name | normalize_name) + "@" + .version
else
    .
end
| if (.purl // "") | startswith("pkg:pypi/") then
    .purl |= sub("^pkg:pypi/(?<name>[^@]+)"; "pkg:pypi/" + (.name | normalize_name))
```

Direct env access (it must be an exported variable, not a shell variable):


```suggestion
jq 'def normalize_name: ascii_downcase | gsub("[-_.]+"; "-");
.dependencies |= map(if .ref | test("^apache-airflow-providers-") then
   .dependsOn |= map(select(. != "apache-airflow==" + env.PRODUCT_VERSION))
else
   .
end)
| .components |= map(if (.purl | not) and (.name != null) and (.version != null) then
    .purl = "pkg:pypi/" + (.name | normalize_name) + "@" + .version
else
    .
end
| if (.purl // "") | startswith("pkg:pypi/") then
    .purl |= sub("^pkg:pypi/(?<name>[^@]+)"; "pkg:pypi/" + (.name | normalize_name))
```

TIL: You can also use these styles:

```
jq -n '$ENV.PRODUCT_VERSION'
jq -n 'env["PRODUCT_VERSION"]'
jq -n 'env.PRODUCT_VERSION'
```

NickLarsenNZ (approved):
LGTM

StefanFl (approved):

airflow/Dockerfile
NickLarsenNZ: ok, github doesn't make that easy to see the difference. Basically example 1 had `jq --arg product_version "$PRODUCT_VERSION" ...` then used `$product_version`.

Example 2 used `env.PRODUCT_VERSION`

PR #1636, opened 2026-09-14
feat(hive): Restore the Thrift methods removed by HIVE-26537 for 4.2.0

lfrancke:
# Description

Hive 4.0.1 removed get_table and get_table_objects_by_name from the metastore Thrift interface. Any client built against a Hive 2.3 metastore client still calls them, and that is what Spark bundles: 2.3.9 in Spark 3.5, 2.3.10 in Spark 4.0 through 4.2. Iceberg's HiveCatalog reaches the metastore through IMetaStoreClient and therefore uses whichever client Spark supplies, so an Iceberg job against a Hive 4.0.1 or newer metastore fails.

This restores the methods as they were in Hive 4.0.0.

This also fixes a related problem: We set the version to build to include the `-stackable` postfix. And we'd expect Maven to use those artifacts it builds _in the same build process_. Well...it didn't because Hive has - in some places - hardcoded the wrong version which means that Hive would not actually use the things we build but instead downloads the dependencies from Maven Central. This fixes that as well.

I ran the Hive kuttl test against this and they all succeeded. That doesn't say much because they don't test the Iceberg behavior which is why there is another PR for hive-operator upcoming that adds a test there.

> [!NOTE]

<details>

```shell
```

</details>


hive/stackable/patches/4.2.0/0005-Resolve-every-in-reactor-dependency-against-the-modu.patch
NickLarsenNZ, line 1: > Hive has - in some places - hardcoded the wrong version which means that Hive would not actually use the things we build but instead downloads the dependencies from Maven Central.

Wow!

Is this something we can/should upstream?

hive/Dockerfile
```diff
@@ -83,9 +83,13 @@ elif [[ "${PRODUCT_VERSION}" == 4.0.* ]]; then
   )
 else
   # Starting with 4.1.0 the build process changed again in https://github.com/apache/hive/pull/5936 (HIVE-29062)
+  # The main reactor pins its dependencies on the standalone-metastore modules through `standalone-metastore.version`.
+  # This is _not_ changed by `mvn versions:set` above.
+  # If we do not change it here it'd download those dependencies (without the -stackableXXXX suffix) from Maven central.
   mvn \
     clean package \
     -Dhadoop.version=${HADOOP_VERSION}-stackable${RELEASE_VERSION} \
+    -Dstandalone-metastore.version=${NEW_VERSION} \
```
NickLarsenNZ, line 92: How did you find this? Did you just see `mvn` output showing dependencies without `-stackable`?

NickLarsenNZ: Is this something we need to do across all other products?

NickLarsenNZ (dismissed):
LGTM

hive/Dockerfile
lfrancke: Claude found it.

It was relatively easy though: We looked at the resulting folder with the jars and found Hive jars that didn't have our `stackable` suffix. All of those should come from "within" the build so we hunted it down.

I had Claude look briefly at our other Dockerfiles and it said they are looking good but I don't think I let it look at the results in the actual images, so it might still be worth it.

hive/stackable/patches/4.2.0/0005-Resolve-every-in-reactor-dependency-against-the-modu.patch
lfrancke: Part of it is already upstreamed.
https://issues.apache.org/jira/browse/HIVE-29827 and some other related work is ongoing https://issues.apache.org/jira/browse/HIVE-30010 which is why I didn't want to open another PR while that second one seems to be in progress. But yes, in theory we should upstream all of this, I had a reason not to though in this case.... should those upstream issues not move we should revisit that.

lfrancke:
Had to fix the changelog conflict.

NickLarsenNZ (approved):
LGTM

PR #1660, opened 2026-09-16
chore(boil): Remove some dead code

Techassi:
We ideally want `dead_code_pub_in_binary = "warn"`, but that is currently nightly only and produces noise when running unit tests. See https://github.com/rust-lang/rust/issues/74970.

For now, we could replace all `pub` items with `pub(crate)` to get proper dead code warnings again.


lfrancke (dismissed):

lfrancke:
You're running into the same rustls issues I did yesterday :(

Techassi:
Hehe yeah, I have been dealing with them yesterday and today as well. I will bump the crate to address the failure.

lfrancke (approved):

